PR #11, opened 2016-08-01
getValue() : return items or selected choices

chymz:
Maybe it can be useful :)


jshjohnson:
Nice addition, thanks

pashkes:
why I get always object of selected value. I need to get array of options

PR #30, opened 2016-08-24
added !default to variables assignment in choices.scss

flamytwista:
This allows to change values when including with webpack from node_modules.
Similar technique is used in twitter bootstrap for example.


jshjohnson:
Nice, thanks

PR #34, opened 2016-08-31
Optional sorting

BroiSatse:
When using Choices as an alternative for those pesky select tags, sorting select's options is not really the feature I'm looking for. This PR allows user to turn the sorting off completely (leaving it on when search is performed).

I have tried other methods of preventing the sorting from happening, without success. See: http://stackoverflow.com/questions/39249946/javascript-bypassing-sort-attempt-yielded-weird-behaviour#39250125


jshjohnson:
Thanks! Nice addition.

PR #38, opened 2016-09-12
Added bower.json file

Ameriks:
I added bower.json file so that this package could be installed more easily with bower.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,21 @@
+{
+  "name": "choices.js",
+  "description": "A vanilla JS customisable text input/select box plugin",
+  "main": [
+      "./assets/scripts/dist/choices.js",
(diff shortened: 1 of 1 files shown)


jshjohnson:
Hmm, is Bower really that popular nowadays? I thought NPM had once that battle!

Ameriks:
Not really. There are still missing good solutions to manage front-end dependencies (or I haven't heard). NPM has browserify, but it doesn't manage css dependencies. I have created bower postinstall script that copies only required files from package and gulp.js packs all JS and (S)CSS. And it works.

btw stats show that bower is still popular enough to not ignore it. :) https://bower.io/stats/

jshjohnson:
Fair enough - thanks!

PR #44, opened 2016-09-26
Add the ability to define "Press to select" text through configuration

froggiefab:
Hi,

This PR aims to define the "Press to select" text through Javascript configuration as any other text.

The "attr()" CSS function has been used. Its support for the "content" property is well supported (https://developer.mozilla.org/en/docs/Web/CSS/attr#Browser_compatibility).

Thanks


jshjohnson:
Awesome, thanks

PR #45, opened 2016-09-26
Compute input size with an 'input' element

froggiefab:
Hi,

It seems that using a 'span' element is not as accurate as using a real 'input' element to compute components input's width.

With a 'span', longer values get truncated wich seems to be solved with this PR.

Thanks


jshjohnson:
Thanks for the PR. Unfortunately this doesn't seem to work for me. As the input element doesn't grow to the size of it's contents, the width doesn't change (see screenshot).

I have just put in a fix that will resolve an input width issue when editing an item however so thanks for prompting me!

Josh

screen shot 2016-09-26 at 13 59 43

Issue #46, opened 2016-09-26
Server side filtering


froggiefab:
Hi,

I'd like to add a server-side filtering feature which would fetch data as the user inputs text (with a 'minChars' and a 'delay' properties). 

Is that something that you plan to add in a future release ?

Thanks

jshjohnson:
Hi @froggiefab,

So just I'm clear, you would like the ability to specify how many characters are needed before Choices will filter as well as the delay after each key press? 

Thanks,

Josh

froggiefab:
Yes that would be great.

But I also would like to be able to fetch choices from the server as the user inputs text (make a request to receive filtered choices from a large dataset rather than loading all choices on init).

Thanks,

Fabien

jshjohnson:
Ah I see what you mean. I could potentially add a new callback for when the user types so you can then handle your server logic from within that callback.

I'll look into this, sounds like a cool idea.

froggiefab:
Great, thanks a lot !

jshjohnson:
As of version 2.1.0 there is now a callback when a user searches + you can now specify a minimum character limit via the [ 'searchFloor'](https://github.com/jshjohnson/Choices#searchfloor) config property.

PR #47, opened 2016-09-27
Server side filtering

froggiefab:
Hi,

This is a PR for the issue #46.

Not sure if this is what you had in mind though.

I think it might be a good idea to relie on async actions for the Redux/Store object. The dependencies needed would probably make Choices.js bigger though.


jshjohnson:
Just so I can test this, could you provide a simple Choices config using this functionality please?

Thanks!

froggiefab:
I just pushed a sample in the 'index.html' file.

jshjohnson:
There are a lot of improvements you've made within this pull request that I will be merging in - I have removed (for now) the ability to live search though, as I found it caused race conditions with Redux - I'll look into some async Redux libraries to see if there are any solutions. I have kept the search callback though so there is still the opportunity of handling async actions when a user searches!

Thanks for all your help!

froggiefab:
Thanks a lot !

FriOne:
@jshjohnson Is this in the last version? I want to create wrapper on angular.

PR #51, opened 2016-10-12
Shorten list style declaration

zslabs:
According to http://stackoverflow.com/questions/19889069/whats-the-difference-between-list-style-type-and-list-style `list-style` can be substituted for `list-style-type` if it's meant to replace all relevant sub-styles (position and image).


jshjohnson:
Thanks!

PR #67, opened 2016-11-01
Makes default settings publicly accessible

Hacktivista:
This allows to change default settings globally. Useful for design and internationalization.


jshjohnson:
Hey @Hacktivista,

What are the benefits of this approach? This would change how you set custom settings right?

Thanks

jshjohnson:
Also seeing that this causes duplications when adding choices via the config.

Hacktivista:
This allows you to change settings before the object creation, changes defaults globally.

How does this causes duplications? Actually I just moved the variable fromminside to outside.

jshjohnson:
I think it causes duplications (and fails some tests) because the defaults do not reset. So if you add three choices via the config and then want three other choices in another config, the second will have choices from both.

PR #72, opened 2016-11-07
Callback calls using 'call' to force 'this'

froggiefab:
Hi,

In order to retrieve the instance of Choices that triggered a callback, the callbacks are now invoked using the 'call' function to explicitly set 'this' inside the callback.

Thanks


jshjohnson:
Hey @froggiefab,

Thanks for this - was this causing a bug for you?

Thanks

froggiefab:
Yes and no :)

I was previously using the "passedInput" parameter of the 'callbackOnChange' function to retrieve the instance.
This parameter was removed so I don't know wich instance triggered the callback anymore.

I thought it would be nice to have a way to retrieve the instance for all triggered callbacks.

jshjohnson:
Makes sense. Thanks - i'll review this ASAP.

PR #79, opened 2016-11-18
Fix item selection against manual input

StephaneTrebel:
Your prototype was working fine except for two issues :

- The automatic highlight of the first list option is conflicting with user input
- Selecting an option with Enter caused adding the current user input (For instance if I input "foo" and the list shows "foobar" and then I hit direction keys to select "foobar" and hit Enter, it would cause to add "foo" as a new item)

I fixed them both in this PR by:
- Making it so the first option is not highlighted by default
- Reversing the order of detecting the onEnter worklow: First we check for highlighted options and only after the user input.

This should fix (https://github.com/jshjohnson/Choices/issues/39)

### DIFF
--- a/assets/scripts/src/choices.js
+++ b/assets/scripts/src/choices.js
@@ -408,7 +408,6 @@ class Choices {
             // If we actually have anything to add to our dropdown
             // append it and highlight the first choice
             this.choiceList.appendChild(choiceListFragment);
-            this._highlightChoice();
           } else {
             const activeItems = this.store.getItemsFilteredByActive();
             const canAddItem = this._canAddItem(activeItems, this.input.value);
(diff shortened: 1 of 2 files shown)


dgt41:
Works great!

jshjohnson:
Love it - thanks for this!

StephaneTrebel:
No problem, I needed it on my side too so I'm glad I could help us both ^^ :+1:

AbdelrhmanS:
حاجة حلوة

AbdelrhmanS:
@dgt41 how this works ? 
Do it want custom configuration ?

dgt41:
@AbdelrhmanS you need to check out https://github.com/jshjohnson/Choices/compare/feature/select-add-items?expand=1
Also you need to add `addItem: true` or something like that in  your config, check that branch's docs

AbdelrhmanS:
i had checkout to this branch #79 and add those config
```javascript
const tagsChoices = new Djaty.libs.Choices(tagsEl, {
   choices: tagsTpl, // my predefined inputs 
   editItems: true,
   maxItemCount: 5,
   removeItemButton: true,
   duplicateItems: false,
   addItem: true,
});
```
```html
<!-- my select tag -->
<select class="tags-input" type="text" placeholder="Tags" multiple></select>
```

PR #81, opened 2016-11-21
Trigger change event on passed element

jbanety:
Adds :
- an utility method to trigger event in plain JS
- two lines in choices.js to trigger change event


jshjohnson:
Thanks for this. 

As I mentioned in the issue, my aim is to keep Choices small in file size. Although this PR only adds a trivial amount of code, to be consistent I would need to trigger events for all callbacks which would add more weight. I also think it would cause confusion in the documentation having 2 callbacks fire for the same event.

jbanety:
Callbacks are not easy to use is many cases :
 - when you create your instances dynamically
 - when you wants to add multiple differents handlers for one event (or callback)
 - when your callback can't be set at initialization time.

You can use a different design with events (like http://www.datchley.name/es6-eventemitter/) and developpers just have to do `myChoicesInstance.addEventListener('removeItem', function(id, value, groupValue) {...`

This is the same design pattern as classical DOM Events in vanilla JS.

jshjohnson:
I really like the look of that event emitter pattern. Might create a feature branch for this.

thejmsill:
I personally really like jbanety's callbacks. It's the classical DOM events

jshjohnson:
I am going to close this pull request as I plan to change callbacks to use the ES6 event emitter instead. Thanks for the inspiration! 👍

Issue #39, opened 2016-09-12
User add item & multi-select


Tam:
Is it possible to mix multi-select with the users ability to add custom items? Or alternatively, give the text field some predefined suggestions in the style of the multi-select?

I've tried setting the `choices` option when targeting a text field, but it doesn't seem to work.

jshjohnson:
The ability to add custom items to a multi-select is a definite possibility - that would be fairly simple to implement. Thanks for the suggestion.

dgt41:
@jshjohnson if you are about to implement this feature can I also ask to apply this also for single select option tag? It would be great help on something I'm currently working...

dgt41:
@jshjohnson I would like to give some input here, so for the single value:

http://codepen.io/dgt41/pen/xRbpwK/

And for the multiple values:

http://codepen.io/dgt41/pen/RoNxro

So couple remarks here: 
- there should be an option for the value (both for sing and multiple)
- the search should return ONLY exact results not partials as it is ATM

Hope this is somehow helpful

dgt41:
@jshjohnson any ETA here?

jshjohnson:
@dgt41 

I'll be looking at this as soon as I can - very busy this week though. I'm happy to review pull requests.

jshjohnson:
Just to give an update, I've begun work on this within [a feature branch](https://github.com/jshjohnson/Choices/tree/feature/select-add-items)

dgt41:
@jshjohnson I did some testing on the feature branch, so for the single the item is added but is not active:
![screen shot 2016-11-14 at 22 48 32](https://cloud.githubusercontent.com/assets/3889375/20286699/dcffefe8-aad0-11e6-8a97-07cff202faf7.png)

For the multiple I do get some strange behaviour, adding a new item will also add another (already existing) item. I hope I can find some time debugging and contributing some code

StephaneTrebel:
FYI, I've a push a PR (https://github.com/jshjohnson/Choices/pull/79) to help complete this feature.
Feel free to comment !

jshjohnson:
Thanks to the PR from @StephaneTrebel, I'm close to getting this feature through the door. 

Things left to do:

- [ ] Ensure the select box stays focussed after inputting a custom choice
- [ ] Ensure the `duplicateItems` config option works with select elements (i.e. no inputting a value that already exists

StephaneTrebel:
FYI, I've done some tests and I think that :

[...]

PR #88, opened 2016-12-14
Feature/select add items when duplicateItems is false

StephaneTrebel:
Hey there !

I did some more testing about issue #39 , especially on `duplicateItems = false` config and I did find out that it was possible to add an existing Choice (and thus, duplicate it).

I changed that to the normal select behaviour which is to add as an Item the existing Choice with same label.


jshjohnson:
Great - thanks! I think this feature is really close to completion. It just needs some tests + some investigation into option groups.

Issue #92, opened 2016-12-29
Empty option values aren't supported


rstacruz:
Can I have empty `value` attributes for options? Seems I can't.

```html
<select name="country">
  <option value="US">United States</option>
  <option value="CA">Canada</option>
  <option value="">None of the above</option>
</select>
```

PR #93, opened 2016-12-29
Allow empty values (#92)

rstacruz:
This allows blank string values, such as:

```html
<select name="country">
  <option value="US">United States</option>
  <option value="CA">Canada</option>
  <option value="">None of the above</option>
</select>
```


jshjohnson:
👌🏻

Issue #97, opened 2017-01-17
Support page up/down to select first/last options


haggen:
Page up and down should jump, respectively, to the first and last options in the dropdown. On macOS it should be command+left/up and command+right/down.

jshjohnson:
Like the idea :) Will be happy to review any pull requests if people want this

haggen:
I'm working on it. Also, `Command`+`Left`/`Right` is equivalent to `Home` and `End` keys, so scratch that, only `Command`+`Up`/`Down` should work.

haggen:
I just regretted that I said I was working on it.

jshjohnson:
Thanks for the PR - in response to your question, I need to know the direction to determine whether we should scroll to the top of the element (scrolling up) or to the bottom of the element (scrolling down). There is probably a more concise way of doing this though...

PR #99, opened 2017-01-25
Accept page up/down to skip to the first/last options

haggen:
#97 

I tried my best not to overcomplicate things and follow the existing code style. Let me know if anything isn't of your liking, I'd be happy to fix it.


jshjohnson:
Thanks 👍

Issue #104, opened 2017-01-30
[BUG] setChoices with groups doesn't honor id in config


mike-robertson:
If you `setChoices` for a list using the group options (array of groups, each which has choices within them), the `id` field passed is not used at all to set the group's id. Instead, the `index` of the group in the config array object is used. This is a problem because if you are trying to add a group at a time, the index is always 0 (so all groups have a groupId of 0). Example:

```js
myChoices.setChoices([{
  label: 'myGroup1',
  id: 1,
  disabled: false,
  choices: choices1Array,
}], 'value', 'label');
myChoices.setChoices([{
  label: 'myGroup2',
  id: 2,
  disabled: false,
  choices: choices2array,
}], 'value', 'label');
```

I'd actually have this as the state:

```js
myChoices.currentState.groups = [
  {
    label: 'myGroup1',
    id: 0,
    disabled: false,
    choices: choices1array
  }, {
    label: 'myGroup2',
    id: 0,
    disabled: false,
    choices: choices2array
  }
]
```

mike-robertson:
The fix is simple:
https://github.com/jshjohnson/Choices/blob/master/assets/scripts/src/choices.js#L793

Instead of just passing `index` there in the id arg, pass 
```js
result.id || index
```
However, I think that doing index there is weird because of the issue I highlighted. It should probably error if there is no id or generate on if none is given, not use the index.

jshjohnson:
Resolved as of [version 2.7.0](https://github.com/jshjohnson/Choices/releases/tag/v2.7.0)!

PR #105, opened 2017-01-30
use result.id instead of the index in all places that do that

mike-robertson:


mike-robertson:
fixes issue #104

jshjohnson:
Looks good! Are you able to put an error in if no group ID is passed?

mike-robertson:
Would you prefer that over generating a uuid (using something like this https://www.npmjs.com/package/uuid)

mike-robertson:
I just threw an error stating that group id is required if it is missing. If you would rather do the UUIDs, it should be pretty simple to implement that instead.

jshjohnson:
I tweaked this slightly to avoid repetition - so if a null ID is passed to `_addGroup`, we generate a unique one.

Thanks for this! 💯

PR #107, opened 2017-01-31
Add default type for remove item buttons

### LINKED ISSUE #103: Remove Item Button has no type
The nice little remove item button has no type, which can cause very unfortunate side-effects on some browsers like Firefox that make their type "submit" by default (see [the button documentation on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type)). I haven't had the issue with Chrome, though.

Undesired side-effects include submitting forms on item removal, for instance, or removing items from choices input upon implicit form validation if no other submit button exist.

I recommend giving them a type="button" to avoid any doubt about their purpose: they are javascript oriented elements and nothing else.

StephaneTrebel:
- fixes #103


jshjohnson:
👌🏻

PR #122, opened 2017-03-01
Feature/label within event add remove item

alexb974:
Hello, 

I thought that sometimes it could be interesting to have the label passed in the events of add/remove items (e.g. The label could be used to be displayed in a message).

I added an example in index.html. 

Thanks.


jshjohnson:
Nice one! Thanks 👌🏻

PR #138, opened 2017-04-04
select-one input is now not focusable in disabled state (fixes #137)

### LINKED ISSUE #137: disable() should make select not focusable
This should prevent element to be focusable. But disabled choice still has ```[tabindex="0"]```
```
const choice = new Choices(element);
choice.disable();
```

I can PR this if you are interested.

adammockor:


jshjohnson:
Awesome, thanks

PR #139, opened 2017-04-05
Feature/select placeholders

adammockor:
Still WIP, do not merge this.


adammockor:
I merge master, I will try to rebase later.

You can try this, placeholder choice should always first choice, regardless of sorting.

Missing thing is that `removeItemButton` should set placeholder after button click, but now it leaves only blank space.

If everything else is missing, let me know.

adammockor:
So I add functionality which set placeholder option after click on remove button item. If there is no placeholder, select will be blank.

Can we get this merged, or is there anything else?

I tried squash commits from master which are hanging here, but without success, because there were lot of merge conflicts and I am not confident about it.

jshjohnson:
Thanks for this, I'll give this a proper review ASAP as it is a major change!

Issue #120, opened 2017-02-27
Feature Request: Ability to disable filter of choices without removing the input


manelmadeira:
Hi,

I would like to request to add a settings flag to disable the choices filtering when typing the search input but continue to emit the search event.

I'm using the on search event to search on an API endpoint.

Thanks,
Manuel Madeira

jshjohnson:
Hey, 

This would be really easy to implement, I would be happy to review a pull request!

Thanks

PR #140, opened 2017-04-06
Add option to remove search filter choices but emit the search event.

manelmadeira:
To solve the issue #120 .

I've fixed the unit test of searching too.


jshjohnson:
Awesome, thanks :D

PR #145, opened 2017-04-11
Export es6 specific version

jshjohnson:


jshjohnson:
Closing this as I don't really think it's necessary - the actual JS files are themselves ES6 versions

PR #153, opened 2017-05-02
Fix/dropdown height

adammockor:
* If you use css transition on dropdown, like scale, `getBoundingClientRect.height` value is scaled. So `offsetHeight` gives accurate height value.
* Do not remove `flippedState` class if `shouldFlip !== true`. This class can be added manually before and should stayed there.

### DIFF
--- a/assets/scripts/src/choices.js
+++ b/assets/scripts/src/choices.js
@@ -635,7 +635,7 @@ class Choices {
     this.dropdown.setAttribute('aria-expanded', 'true');
 
     const dimensions = this.dropdown.getBoundingClientRect();
-    const dropdownPos = Math.ceil(dimensions.top + window.scrollY + dimensions.height);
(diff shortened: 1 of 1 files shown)


assets/scripts/src/choices.js
```diff
@@ -647,8 +647,6 @@ class Choices {
 
     if (shouldFlip) {
       this.containerOuter.classList.add(this.config.classNames.flippedState);
-    } else {
-      this.containerOuter.classList.remove(this.config.classNames.flippedState);
```
jshjohnson: Hmm, why would we keep the flipped state class if the dropdown shouldn't flip?

adammockor: I was adding `is-flipped` class dynamically after choices init in certain scenarios of mine. And then `showDropdown` method was removing this class, because there was enough space.

But I can remove it, if you want and do it with different class on my own.

adammockor: That is probably better aproach anyways, but I am still not sure why to remove this class in `showDropdown`.

PR #155, opened 2017-05-04
Compute placeholder from getComputedStyle after input is mounted to DOM

### LINKED ISSUE #154: Placeholder width
I got problems with computing placeholder width. See image below.

screenshot 2017-05-03 16 47 56

Problem is in `getWidthOfInput` function returning wrong value, because it's getting width from span attached directly to body element, not inside of choices container.

I can imagine 2 different solutions to this problem. Both can be implemented simultaniously.

1. Expose `setInputWidth(width)` to userland with optional width parameter, which can override computed value. So user can set width manualy after `callbackOnInit` for example.

2. Call `_setInputWidth` after input is mounted to DOM (now it's called before mounting in `_createInput`). When input is mounted we can get actual css with `window.getComputedStyle(this.input)` and apply to `testEl`. [Somthing like here](https://github.com/JedWatson/react-input-autosize/blob/master/src/AutosizeInput.js#L71).

I can PR this, but please tell me @jshjohnson which solution you prefer, or if you can think of something different.

adammockor:
Fixes #154

### DIFF
--- a/assets/scripts/src/choices.js
+++ b/assets/scripts/src/choices.js
@@ -2316,7 +2316,6 @@ class Choices {
     if (this.placeholder) {
       if (this.passedElement.type !== 'select-one' && !this.config.searchPlaceholderValue) {
         input.placeholder = this.placeholder;
-        input.style.width = getWidthOfInput(input);
       } else {
         const placeholderItem = this._getTemplate('placeholder', this.placeholder);
         this.itemList.appendChild(placeholderItem);
@@ -2346,6 +2345,14 @@ class Choices {
       containerInner.appendChild(input);
     } else if (this.canSearch) {
       dropdown.insertBefore(input, dropdown.firstChild);
+    }
+
+    if (
+      this.placeholder &&
+      this.passedElement.type !== 'select-one' &&
+      !this.config.searchPlaceholderValue
+    ) {
+      input.style.width = getWidthOfInput(input);
     }
 
     if (this.isSelectElement) {
--- a/assets/scripts/src/lib/utils.js
+++ b/assets/scripts/src/lib/utils.js
@@ -477,6 +477,19 @@ export const getWidthOfInput = (input) => {
     testEl.style.width = 'auto';
     testEl.style.whiteSpace = 'pre';
 
+    if (document.body.contains(input) && window.getComputedStyle) {
+      const inputStyle = window.getComputedStyle(input);
+
+      if (inputStyle) {
+        testEl.style.fontSize = inputStyle.fontSize;
+        testEl.style.fontFamily = inputStyle.fontFamily;
+        testEl.style.fontWeight = inputStyle.fontWeight;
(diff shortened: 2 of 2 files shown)


assets/scripts/src/lib/utils.js
```diff
@@ -477,6 +477,19 @@ export const getWidthOfInput = (input) => {
     testEl.style.width = 'auto';
     testEl.style.whiteSpace = 'pre';
 
+    if (document.body.contains(input) && window.getComputedStyle) {
+      const inputStyle = window.getComputedStyle(input);
```
adammockor: This is working on your example page as intended, if you set `<style>.choices__input
{ font-size: 20px; }</style>`. 
But it's not working on my own codebase. `getComputedStyle` gets defaults, instead of actual values. I have to do it in async e.g. in `setTimeout`, but I don't know why.

assets/scripts/src/choices.js
```diff
@@ -2348,6 +2347,14 @@ class Choices {
       dropdown.insertBefore(input, dropdown.firstChild);
     }
 
+    if (
+      this.placeholder &&
+      this.passedElement.type !== 'select-one' &&
+      !this.config.searchPlaceholderValue
+    ) {
+      input.style.width = getWidthOfInput(input);
```
adammockor: This works for my codebase.
```
setTimeout(() => {
  input.style.width = getWidthOfInput(input);
});
```

jshjohnson: Hmm, this isn't ideal. What is `getWidthOfInput` actually returning?

adammockor: Imagine I got three selects with different sizing `.select--large`, `.select--small` and without class for base size. Without `setTimeout` delay it always returns styling of base select size. With `setTimeout` it returns correct values. It looks like that css applies slower in larger codebase and we need to wait little bit for that.

jshjohnson: I wonder if we could achieve the same thing with `requestAnimationFrame()` - as we probably only need to wait 1 frame

adammockor: Yes, it also works with requestAnimationFrame.

adammockor: So maybe can we do this?

jshjohnson: Interesting, when I tested this I didn't need to use `requestAnimationFrame` nor `setTimeout`. Hmmmm

adammockor: Yes, me neither in this project. But in my larger project it won't work without it. But I am ok to leave as it is now without async call and do it myself in `callbackOnInit`.

jshjohnson: Are you loading Choices asynchronously? Wonder whether that's the reason

adammockor: No I am not. I meant just `getWidthOfInput(input);` calling asynchronously.

PR #160, opened 2017-05-16
Fix dropdown hiding on blur in IE11

rstacruz:
Fixes an issue where clicking on a dropdown scrollbar will close that dropdown in MSIE 11.

## Steps to reproduce

Make a Choices instance of `select-one` type, with a search bar, with enough items to have a search bar in the dropdown.

image

When you click on the scrollbar in IE, it will close the dropdown because it fires the `blur` event on the text field.

## Fix

This closes the dropdown when something else in the document receives a focus (`document.addEventListener('focus', ...)`), INSTEAD of when the text field is blurred.

### DIFF
--- a/assets/scripts/src/choices.js
+++ b/assets/scripts/src/choices.js
@@ -1271,6 +1271,7 @@ class Choices {
     document.addEventListener('touchend', this._onTouchEnd);
     document.addEventListener('mousedown', this._onMouseDown);
     document.addEventListener('mouseover', this._onMouseOver);
+    document.addEventListener('focus', this._onDocumentFocus);
 
     if (this.passedElement.type && this.passedElement.type === 'select-one') {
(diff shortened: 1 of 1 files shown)


jshjohnson:
Awesome, thanks

jshjohnson:
I've had to revert this change as the document focus event does not seem to fire in all browsers. In Chrome for instance, tabbing through each instance would not close the drop down.

minajevs:
Any updates on this? Are there any workarounds to fix this issue?

minajevs:
Any updates on this? Are there any workarounds to fix this issue?

jshjohnson:
This has now been resolved as of version [2.8.12](https://github.com/jshjohnson/Choices/releases/tag/v2.8.12).

PR #175, opened 2017-05-30
Check `item.value` is type 'string' before calling `trim()`

adamsro:
When selecting a choice with an integer value which has been added via the `setChoices` method I get the following error: `Uncaught TypeError: value.trim is not a function`

I think this PR would be the cleanest fix?


jshjohnson:
Good spot, thanks! 👍

adamsro:
nop. thanks for this lib!

PR #179, opened 2017-06-02
Accessiblity improvements

p-bernal:
- Usage of aria-activedescendant to read options: required setting up ids for each choice
- Allow keys to browse options when screen reader is on by using combobox and listbox roles
- Fix for aria-selected not being set to true when the choices are highlighted
- Fix for text input in searchEnabled mode not being able to get focus anymore after using the arrow keys, closing the dropdown and opening it again.


jshjohnson:
Awesome 💯

PR #181, opened 2017-06-14
fix #94 by adding searchInputAutofocus option

### LINKED ISSUE #94: Allow to disable autofocus for search input
Hello, it is not convenient when the phone keypad appears when you open the drop-down menu.

Now there isn't option to disable autofocus, only if disable search completely, can you add global option to setup "focusInput" (false/true)

    if (focusInput && this.canSearch && document.activeElement !== this.input) {
      this.input.focus();
    }

Thx

MatTheCat:
Attempt to fix #94

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ Or include Choices directly:
     searchFloor: 1,
     searchResultLimit: 4,
     searchFields: ['label', 'value'],
+    searchInputAutofocus: true,
     position: 'auto',
     resetScrollPosition: true,
     regexFilter: null,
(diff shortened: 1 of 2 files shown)


jshjohnson:
Hey @MatTheCat,

I think this is close but it has some issues when tabbing back through inputs. Dropdown's aren't closing when they should be:

screen shot 2017-06-27 at 11 58 08

jshjohnson:
Closing this as focusing/blurring the input is how Choices determines an active dropdown.

Issue #185, opened 2017-06-22
Custom properties on selected items


JayKid:
Hi @jshjohnson,

Thanks for this awesome library, especially when it comes to the accessibility features 😍

However, when using the multiple-select with VoiceOver and going over the already selected items it just reads out "Remove Items" n times but not to what item the buttons actually correspond to.

This could be solved by creating a custom template with `aria-label="Remove item ${data.value}"` but one might need other data to use in the custom templates. The flexibility is there in the template but not when adding an item/choice as the reducer will handpick the attributes, not allowing for custom defined ones to go into the item/choice state.

I want to add other accessibility related attributes. I could create a PR that extends the reducers to take a `customProperties` attribute of the item that aggregates all properties defined by the user in an object so that they don't mix with the library ones.

What would you think?

jshjohnson:
Hey @JayKid - this sounds like a great idea! Will happily review a PR

JayKid:
@jshjohnson I'm on it :)

jshjohnson:
FYI - as of version 2.8.5, the value is passed to an aria label like you suggested :)

jshjohnson:
This has now been merged into version [2.8.6](https://github.com/jshjohnson/Choices/releases/tag/v2.8.6) 💯

PR #191, opened 2017-06-28
Custom properties on choices and selected items

JayKid:
PR for issue #185


jshjohnson:
Perfect!! Thanks so much, this is a great addition.

epaezrubio:
How about passing the customProperties as a data-attribute from the markup? @jshjohnson @JayKid

PR #197, opened 2017-06-30
Fix extra parameter in call to _addItem

JayKid:
I introduced one parameter too much in a call to _addItem from _addChoice (here: 0b3cf1a#diff-27edd04b91c69d5ee4cfac4012766ef7R2202) so the former doesn't get the custom properties and can't set them in the newly created item.

Related to issue https://github.com/jshjohnson/Choices/issues/196


jshjohnson:
Thanks :D 💯

PR #206, opened 2017-07-13
Various minor corrections.

maximmig:
ESLint, etc.


assets/scripts/src/choices.js
```diff
@@ -282,9 +281,9 @@ class Choices {
     // Reinstate passed element
     this.passedElement.classList.remove(this.config.classNames.input, this.config.classNames.hiddenState);
     this.passedElement.removeAttribute('tabindex');
-    this.passedElement.removeAttribute('style', 'display:none;');
+    this.passedElement.removeAttribute('style');
```
jshjohnson: Will this not remove all styles? I can't remember off the top of my head whether we had more!

maximmig: It will. I automatically corrected this line, not thinking much. But your code `.setAttribute('style', 'display:none;')` erases all the styles previously set on the passed element. Need to investigate, thanks!

maximmig:
Here we go, tried it here: https://jsfiddle.net/maximmig/7dvpzo5f/

maximmig:
And again: persisting nulls is a bad idea.

jshjohnson:
Thanks!

PR #209, opened 2017-07-17
Backup & recover original styles: added test

maximmig:
I should've done this earlier, better late than never.


jshjohnson:
Haha, thanks for fixing all my typos ;)

PR #214, opened 2017-07-20
Add MDN's Object.assign polyfill to the ones imported by the choices library

JayKid:
Problem: Object.assign is not available in IE and older browsers 

Solution: Add polyfill from MDN

Fixes https://github.com/jshjohnson/Choices/issues/213


jshjohnson:
I included the majority of polyfills needed to make Choices work down to low versions of IE but many people highlighted that polyfills should be something the developer should manage rather than the library. In the future, I'm aiming to not include any polyfills at all for this reason!

Thanks

PR #216, opened 2017-07-21
Added 'scrollToEnd' event

maximmig:


jshjohnson:
Hey @maximmig,

Although this is cool, I feel like it's something somebody could write themselves if they needed this functionality - I'm hesitant in adding features people haven't asked for so I can keep file size down :)

Thanks though, and if somebody else requests this, I'll re-open this PR. 

Josh

maximmig:
This could help to load data via AJAX page by page. It's a rather simple addition, so why not to have it? Since the component cannot do it, I mean partial data loading, this event could shorten the amount of the boilerplate code that is needed to be written to achieve it.

vkrol:
+1

edtownend:
I could really use this feature so I can implement pagination, it's the only thing keeping me on select2.

PR #218, opened 2017-07-27
Add types

zackschuster:
(Documentation taken from the README, except where source code disagrees or adds data.)

These are the types I'm using for a separate project using TypeScript. They're proven to work in dev and prod using a webpack setup.

Thanks for the great library!


zackschuster:
Updated for v3.0.0. All changes taken from the README.

jshjohnson:
Cool - thanks. Is there a way of testing this config so that it doesn't go out of date?

zackschuster:
I could add a TS file to cover compiler and runtime checks. Fairly barebones stuff, the worst of which would be testing methods. On the flipside you'd have to add typescript as a devDep, which adds ~25mb of weight to your `node_modules`.

PR #229, opened 2017-08-16
Add aria-activedescendant attribute only when dropdown is active

vkolmakov1:
Add `aria-activedescendant` attribute only when an item is highlighted and the dropdown is active + remove it when the dropdown closes.

The attribute (when set on the `input`) originally prevented virtual keyboard from appearing on tablets with IE11 (5eac8d4)

Adding the attribute conditionally (and removing it as soon as dropdown closes) resolves the problem with virtual keyboard in IE11 (tested on a Surface with IE11) while still keeping the required links for screen readers.

### DIFF
--- a/assets/scripts/src/choices.js
+++ b/assets/scripts/src/choices.js
@@ -797,6 +797,10 @@ class Choices {
     this.containerOuter.setAttribute('aria-expanded', 'false');
     this.dropdown.classList.remove(this.config.classNames.activeState);
     this.dropdown.setAttribute('aria-expanded', 'false');
+    // IE11 ignores aria-label and blocks virtual keyboard
+    // if aria-activedescendant is set without a dropdown
+    this.input.removeAttribute('aria-activedescendant');
+    this.containerOuter.removeAttribute('aria-activedescendant');
 
     if (isFlipped) {
       this.containerOuter.classList.remove(this.config.classNames.flippedState);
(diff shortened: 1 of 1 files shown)


jshjohnson:
Thanks! 👍

assets/scripts/src/choices.js
```diff
@@ -2198,7 +2202,16 @@ class Choices {
       // Highlight given option, and set accessiblity attributes
       passedEl.classList.add(this.config.classNames.highlightedState);
       passedEl.setAttribute('aria-selected', 'true');
-      this.containerOuter.setAttribute('aria-activedescendant', passedEl.id);
+
+      const hasActiveDropdown = this.dropdown.classList.contains(
+        this.config.classNames.activeState
+      );
+      if (hasActiveDropdown) {
```
jshjohnson: Quick question - is there a reason why we are setting the aria role inside `_highlightChoice` rather than `showDropdown()`?

vkolmakov1: That is something I was wondering about as well, it would be nice to add the links as soon as dropdown is open. This way the currently highlighted option will be read immediately.

`_highlightChoice` seemed like a good place to update the links because it ensures that they stay up-to-date with mouse hover/arrow key press events.

I've also tried to add a call to `_highlightChoice` at the end of `showDropdown`, which updates the links properly, but prevents virtual keyboard from appearing in IE11. However, the keyboard appears as expected if we call `_highlightChoice` on a next tick (e.g., a timeout of 0). That approach ensures that links are added as soon as dropdown is open and keeps the virtual keyboard in IE11, but seems a bit hacky. I am wondering if there might be a better place to call `_highlightChoice` or set the links when the dropdown is open.

Let me know what are your thoughts on that.

ObviouslySorna: @jshjohnson have you had time to consider what @vkolmakov1 is talking about, here?
What's your opinion?

PR #236, opened 2017-08-23
Performance fix for large for large result set

### LINKED ISSUE #173: Searching using ajax can be really slow
I've discovered a performance issue when the ajax method returns lots of results and you call it repeatedly. This is caused by the loop at [line 1232 of choices.js](https://github.com/jshjohnson/Choices/blob/master/assets/scripts/src/choices.js#L1232).

For example when the ajax method returns 100 results and is called 2 times, the render method is called 200 times! I fixed this issue by implementing a simple async render method using requestanimationframe. This works fine for my use case but it might break backwards compatibility when the user queries the dom right after calling a method on the choices instance.

@jshjohnson Are you interested in a pr?

Remcoman:
Rendering now occurs after one animation frame to prevent spamming of the render function.
You can still call render manually to cancel the scheduled rendering.

This should resolve #173

### DIFF
--- a/assets/scripts/src/choices.js
+++ b/assets/scripts/src/choices.js
@@ -140,7 +140,7 @@ class Choices {
     }
 
     // Create data store
-    this.store = new Store(this._scheduleRender.bind(this));
+    this.store = new Store();
 
     // State tracking
     this.initialised = false;
@@ -202,6 +202,7 @@ class Choices {
 
     // Bind methods
     this.render = this.render.bind(this);
+    this._scheduleRender = this._scheduleRender.bind(this);
 
     // Bind event handlers
     this._onFocus = this._onFocus.bind(this);
@@ -263,7 +264,7 @@ class Choices {
     // Generate input markup
     this._createInput();
     // Subscribe store to render method
-    this.store.subscribe(this.render);
+    this.store.subscribe(this._scheduleRender);
     // Render any items
     this.render();
     // Trigger event listeners
@@ -1950,11 +1951,6 @@ class Choices {
 
     // If target is something that concerns us
     if (this.containerOuter.contains(target)) {
-      // Handle button delete
-      if (target.hasAttribute('data-button')) {
-        this._handleButtonAction(activeItems, target);
-      }
-
       if (!hasActiveDropdown) {
         if (this.isTextElement) {
           if (document.activeElement !== this.input) {
--- a/assets/scripts/src/choices.js
+++ b/assets/scripts/src/choices.js
@@ -140,7 +140,7 @@ class Choices {
     }
 
     // Create data store
-    this.store = new Store(this.render);
+    this.store = new Store(this._scheduleRender.bind(this));
 
     // State tracking
     this.initialised = false;
@@ -175,6 +175,7 @@ class Choices {
 
     this.highlightPosition = 0;
     this.canSearch = this.config.searchEnabled;
+    this.renderScheduled = 0;
 
     this.placeholder = false;
     if (!this.isSelectOneElement) {
@@ -479,6 +480,9 @@ class Choices {
    * @private
    */
   render() {
+    // cancel any scheduled render
+    this._cancelScheduledRender();
+
     this.currentState = this.store.getState();
 
     // Only render if our state has actually changed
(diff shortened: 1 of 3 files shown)


Vercjames:
I made a fork showcasing the lockup I'm experiencing when calling large JSON objects.
-- https://github.com/vercjames/choices

I'm fetching car models based on their make. The only thing I altered from the jshjohnson's master branch was the index.html where I replaced the API and removed the components I'm not using. 

There are 2 API's so please toggle between the 2 to experience the issue yourself. One is Toyota where I pull in 53 records and takes about 1 maybe 2 seconds to load. The other is Honda where I pull in 788 records and it freezes the browser for about 20 seconds. (You know the browser is locked because you cant click any of the other dropdown's)

I tried your change as well Remcoman and recompiled the the dist directory (choices.min.js) and didnt see a difference. Am I missing something?

Remcoman:
I've pushed some new fixes which seem to work on my machine. However this causes all tests to mysteriously fail with the error: "null is not an object (evaluating 'templates[template]')".

This probably has something to do with the use of requestAnimationFrame. Does anyone know of a fix for this problem?

Remcoman:
Hi Vercjames, i pushed some new commits. Could you try to test again?

Vercjames:
Compiled, tested, and working! 3000 records load in under 2 seconds and it doesn't lock up the browser!

assets/scripts/src/choices.js
```diff
@@ -1921,11 +1953,6 @@ class Choices {
 
     // If target is something that concerns us
     if (this.containerOuter.contains(target)) {
-      // Handle button delete
```
jshjohnson: Hmm, why is this being removed? 🤔

Remcoman: This code block threw an error because the delete is also handled in the onMouseDown. Removing this code block did not seem to cause errors (i could still click the delete button).

jshjohnson: I have a funny feeling it's there for when somebody is on a touch device. It may just be unnecessary code though!

Vercjames:
So is this change slated for implementation? or is the slow AJAX rendering still an issue that needs to be resolved?

jshjohnson:
This will be merged in but I'm in the process of a mini-rewrite so am not merging new PRs into master. If you could do me a big favour, it would be be great if you could change this PR to merge into `feature/modularise-code` instead.

Thanks

ArielGueta:
Wow!! My browser was blocked with 300 items and this solved my problem! It works very fast with 900 items. Please merge this as soon as possible.

wojciechczerniak:
@Remcoman @jshjohnson Since this moves `render()` to `requestAnimationFrame` callback updates are now async. `Change` event is called but the `element.value` is not there yet or may not be. Same for `hideDropodown`. Validation once called on `change` or `hideDropdown` is unreliable after this patch.

jshjohnson:
You have some conflicts 👍

MaksimShakavin:
@jshjohnson,
As @wojciechczerniak mentioned, the issue here is not only in the conflicts. Callbacks here are not working properly anymore. All the logic of the library, that depends on the synchronous style of render() method also, i.e destroy method.
Btw, for me this is the most critical issue of the library, that makes it unusable with ajax in IE at all

Aiso786:
Hello guys, when do you think this is going to be included in the master ?

Than you very much for your work.

travist:
The solution to this problem does not have to be so complicated. Please see https://github.com/jshjohnson/Choices/pull/299. I saw 100x performance boost with my solution and did not require complicated threading and polyfills.

I hope we can get a release out soon that incorporates this change.

jshjohnson:
@travist Does your solution alleviate the issue @wojciechczerniak mentioned?

travist:
It does not, but the problem is pretty serious. 

The real issue is within the ```setChoices``` method which calls ```addChoice``` https://github.com/jshjohnson/Choices/blob/master/assets/scripts/src/choices.js#L1001. This could be solved if we just pass in a boolean to say that the adding of that choice is ```async``` and will not trigger multiple ```render``` calls for every item that is added to the list. 

I took a look at this, but am not familiar with the whole reducer thing you have going on in this library and cannot determine how the ```render``` method is re-triggered for every choice that is added. I will try to find some time to debug this, but for now, I just hacked our library that depends on this library to solve it in an async way.

https://github.com/formio/formio.js/blob/master/src/components/select/Select.js#L13

So far from our testing, there is not any negative effects from this.

jshjohnson:
Closing in favour of https://github.com/jshjohnson/Choices/pull/310

PR #241, opened 2017-08-30
Allow user to set his own defaults (simpler)

Hacktivista:
Makes little sense to set the same defaults a lot of times (such as phrases and styles), so user should be able to set his own once.

Related to [PR #67](https://github.com/jshjohnson/Choices/pull/67/commits/f5c077f6f25a211194948a7bbe5fe3da686451fa). More idiomatical (but bulkier) version in [PR #240](https://github.com/jshjohnson/Choices/pull/240)

### DIFF
--- a/assets/scripts/src/choices.js
+++ b/assets/scripts/src/choices.js
@@ -128,7 +128,7 @@ class Choices {
     };
 
     // Merge options with user options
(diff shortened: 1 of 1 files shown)


jshjohnson:
Cool, thanks 👍

PR #244, opened 2017-09-05
Change window.scrollY to window.pageYOffset for IE

MaksimShakavin:
## This is the problem:
IE do not support `window.scrollY`, so flipping is not working 
## This is my solution:
Change `window.scrollY` to more cross-browser  `window.pageYOffset`

Compatibility tables:
https://developer.mozilla.org/en-US/docs/Web/API/Window/pageYOffset
https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollY


jshjohnson:
Cool, thanks 👍

PR #245, opened 2017-09-05
Fixed typescript typings

### LINKED ISSUE #243: Why is the TypeScript definition file not included in the NPM package?
As the title says. I am adding `choices.js` `^3.0.2` to my package json but the typescript definition file is missing. I can see it exists on this Github. Currently I have to manually add the TS defintinion file to my project and include a reference to that in my script. 

I would like the node package manager to actually include the typescript definition file. The `package.json` seems to point to the typescript definition file, so I assume the file is not supposed to be missing from the package.

yannicsmeets:
index.d.ts will now be included in the npm download. Fixes #243.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "3.0.2",
   "description": "A vanilla JS customisable text input/select box plugin",
   "main": "./assets/scripts/dist/choices.min.js",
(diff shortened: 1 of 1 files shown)


jshjohnson:
Perfect, thanks 👍

PR #248, opened 2017-09-09
Adding a placeholder element for single select dropdown support.

travist:
This pull request provides a pseudo-placeholder for single dropdown selects.

Placeholder

### DIFF
--- a/assets/scripts/src/choices.js
+++ b/assets/scripts/src/choices.js
@@ -2256,6 +2256,10 @@ class Choices {
       passedValue += this.config.appendValue.toString();
     }
 
+    if (this.placeholderElement) {
+      this.placeholderElement.style.visibility = 'hidden';
+    }
+
     this.store.dispatch(
       addItem(
         passedValue,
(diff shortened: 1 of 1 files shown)


jshjohnson:
Hey @travist - I'm a bit confused as to what this is resolving :) Could you explain please?

travist:
We use this library for the Form.io javascript renderer @ https://github.com/formio/formio.js. We need to have placeholders be consistent between multiple selects and single select dropdowns and also show consistently across different field types. The way this library deals with placeholders from a technical perspective, where it adds the placeholder property to the select dropdown, is correct, but the display of the dropdown does not show the placeholder. Here is a before and after of our renderer without this fix and with this fix.

**Without Fix** - https://jsfiddle.net/r5Lcw9uq/

**With Fix** - https://jsfiddle.net/ua6zbLvk/

This makes placeholders consistent with Choices.js.  It may be possible I am implementing it wrong, but I have played with every combination of the placeholder settings and could not get it to work like the "fixed" jsfiddle above without this change.

jshjohnson:
Hmm I think this could be achieved by simply adding a placeholder option though, no?

See the default example for a single select [here](https://joshuajohnson.co.uk/Choices/) :)

jshjohnson:
Closing due to inactivity

PR #261, opened 2017-09-21
Update Fuse dependency

JayKid:
Only change needed: Accessing the id property within the Fuse results instead of item.id since Fuse has flattened its object structure and now everything is at top level.

Fixes https://github.com/jshjohnson/Choices/issues/249


jshjohnson:
Hi @JayKid,

I have changed the base on this PR to `feature/modularise-code` as this will be the next release. Would you mind just checking the conflict please?

Thanks

JayKid:
@jshjohnson should be okay now : )

jshjohnson:
👍

PR #269, opened 2017-10-03
Make select one placeholder consistent with other types

brennemankyle:
Remove need for placeholder <option placeholder> element for select one

This pull request increases consistency for users and removes more code than it adds. This change is very useful for frameworks (ng1, ng2, emberjs, reactjs) that check models and html differences. If you have a <option placeholder> element and then added the other choices later, the choices model has to update to add the placeholder, which triggers change cycles. This makes things slow and harder to develop.


jshjohnson:
Hey @brennemankyle,

I'm a bit confused. How would I go about having a placeholder with a select-one after this change?

Thanks

brennemankyle:
Either with an attribute
`placeholder="A placeholder"`
or like other types with the choice config
```
      new Choices('#choices', {
        placeholder: true,
        placeholderValue: 'Select...'
      });
```

ableeda:
Great feature! Let's merge it.

jshjohnson:
You have some conflicts 👍

brennemankyle:
I'll take a look

brennemankyle:
Updated here, now using develop branch: https://github.com/jshjohnson/Choices/pull/294

Issue #270, opened 2017-10-05
HTMLFormElement.reset() does not resets value of Choices element


tekord:
I have a filter form with native Select element and Choices element. `HTMLFormElement.reset()` does not affect Choises elements. I assume it is a bug.

jshjohnson:
Hi @tekord,

This isn't so much a bug - it's more of a missing feature! :)

Aterbonus:
Hi. I'm searching projects for Hacktoberfest and I want to tackle this issue :)

jshjohnson:
@Aterbonus Awesome :D

PR #273, opened 2017-10-09
Resolve issue #270

Aterbonus:
Now Choice elements reacts to ```HTMLFormElement.reset()``` adequately.

I needed to do a deep copy of the state because some reducers modify it causing side effects.

For now, the commits lacks the test for the new feature because I'm not familiar with test suites.


jshjohnson:
You have some conflicts 👍

PR #274, opened 2017-10-10
add canUseDom

SampsonCrowley:
added CanUseDom checking so that it doesn't run for SSR


jshjohnson:
Thanks for this.

Surely there would only need to be conditionals inside the constructors of each class, no? As if they return false, none of their methods will be called? 

Also, the next release will be from `feature/modularise-code`, so would you mind changing he base branch to that instead of master please :)

SampsonCrowley:
I started with just the constructors and still ended up with errors.
Further research is necessary, but this works for the short term and I was
in a hurry to get it up and running

SampsonCrowley:
I'll make a cleaner version from the feature/modularize-code branch later today

jshjohnson:
Closing based on previous comment :)

SampsonCrowley:
sorry for the delay. been really busy. I'll try to get to this by the end of the week

PR #276, opened 2017-10-12
Move icons to css with encoded data

indacloud:
## This is the problem:
the `$choices-button-icon-path` needed to be changed for use with Weppack and I still had some complications when building my project.

## This is my solution:
The icons can be encoded and used directly in the stylesheet. This makes it easier to customize and removes 2 dependencies.
You can then customize `$choices-icon-cross` and `$choices-icon-cross-inverse` with one of your one data or asset.


jshjohnson:
You have some conflicts 👍

PR #204, opened 2017-07-12
Update to Webpack 3

razh:
I rebased #145 on top of master and combined the changes with https://github.com/jshjohnson/Choices/tree/feature/webpack-2.

This adds [`webpack.optimize.ModuleConcatenationPlugin`](https://medium.com/webpack/webpack-3-official-release-15fd2dd8f07b) which results in a smaller bundle size when combined with tree-shaking.

This does reintroduce the unnecessary ES6 bundles, so those could be removed if desired.

| File size | Before | After |
|---|---|---|
| `choices.js` | [186 KB](https://github.com/jshjohnson/Choices/blob/v2.8.7/assets/scripts/dist/choices.js) | [181 KB](https://github.com/razh/Choices/blob/feature/webpack-3-dist/assets/scripts/dist/choices.js)
| `choices.min.js` | [65.4 KB](https://github.com/jshjohnson/Choices/blob/v2.8.7/assets/scripts/dist/choices.min.js) | [58.7 KB](https://github.com/razh/Choices/blob/feature/webpack-3-dist/assets/scripts/dist/choices.min.js) |
| `choices.min.js` (gzip)* | [17.8 KB](https://raw.githubusercontent.com/jshjohnson/Choices/v2.8.7/assets/scripts/dist/choices.min.js) | [16.5 KB](https://raw.githubusercontent.com/razh/Choices/feature/webpack-3-dist/assets/scripts/dist/choices.min.js) |

* I'm using the Network tab in Chrome developer tools as a rough estimate of this.

I've skipped including the build files in this pull request, but I do have a branch with them: https://github.com/razh/Choices/tree/feature/webpack-3-dist

PR #282, opened 2017-11-07
Update to Webpack 3

razh:
Redo of #204 

This adds [`webpack.optimize.ModuleConcatenationPlugin`](https://medium.com/webpack/webpack-3-official-release-15fd2dd8f07b) which results in a smaller bundle size when combined with tree-shaking.

While compiling with `webpack.config.prod.js`, Babel's module transpilation is disabled so that Webpack can tree-shake.


jshjohnson:
Thanks 👍 💯

PR #283, opened 2017-11-07
Add searchEnabled condition for showing dropdown on typing

PetrBulanek:


jshjohnson:
Hmmm, is this actually necessary?

PetrBulanek:
I think so, because when the searchEnabled is set to false and user has focus on Choices element, it's not expected behavior to show the dropdown on any keypress, at least in my opinion. It should probably open on hitting enter, space or down arrow, but not on other random keys. What do you think?

jshjohnson:
Closing as I don't think is necessary 👍 thanks

PR #299, opened 2017-12-02
Resolved performance problems with large datasets.

travist:
I noticed that with large datasets, this library breaks down really bad. After some investigation, I realized that you are "re"-rendering the entire list for EVERY item that is added to the list. This creates a runaway exponential time to render for each new item that is added to the list.

The solution is simple. Add a debounce around the render() method and performance goes up about 10x for large datasets (> 1000 records).

[JSFiddle Before Fix](https://jsfiddle.net/travistidwell/2y03uw13/)  ~1600 records takes more than 30 seconds to render

[JSFiddle After Fix](https://jsfiddle.net/travistidwell/mejrLpu4/)  ~1600 records takes about 2 seconds

### DIFF
--- a/src/scripts/src/choices.js
+++ b/src/scripts/src/choices.js
@@ -70,6 +70,7 @@ class Choices {
     this.currentState = {};
     this.prevState = {};
     this.currentValue = '';
+    this.renderDebounce = null;
 
     // Retrieve triggering element (i.e. element with 'data-choice' trigger)
     const passedElement = isType('String', element) ? document.querySelector(element) : element;
@@ -423,11 +424,22 @@ class Choices {
   }
 
   /**
+   * Render DOM with values.
+   */
+  render() {
+    if (this.renderDebounce) {
+      clearTimeout(this.renderDebounce);
+    }
+
+    this.renderDebounce = setTimeout(() => this._render(), 100);
+  }
+
+  /**
    * Render DOM with values
    * @return
    * @private
    */
-  render() {
+  _render() {
     this.currentState = this.store.getState();
     const stateChanged = (
       this.currentState.choices !== this.prevState.choices ||


jshjohnson:
I feel like requestAnimationFrame would be more suitable here instead of setTimeout. I'm concerned there will still be the issues mentioned [here](https://github.com/jshjohnson/Choices/pull/236#issuecomment-335646250) with this change...

travist:
That concern is still a problem. I gave a response to that thread if you want to read through it. I will try to find some time to debug through the code, but this problem could be solved by passing an ```aync``` flag to the ```addChoice``` method, which will trigger a debounced render.  You could then add a new method called ```renderAsync``` and the code would be changed as follows.

```js
/**
   * Render DOM with values asynchronously with a debounce.
   */
  renderAsync() {
    if (this.renderDebounce) {
      clearTimeout(this.renderDebounce);
    }

    this.renderDebounce = setTimeout(() => this.render(), 100);
  }
```

travist:
On a separate note, I had to add this to our library to get through this problem.

https://github.com/formio/formio.js/blob/master/src/components/select/Select.js#L13

This is for the JSON form renderer @ http://formio.github.io/formio.js/.  So far with the change I made (which is to duck-punch the render method to make it debounced), we have not seen any negative effects.

evu:
This is a major improvement and could do with being merged asap, however I noticed the setValueByChoice no longer works: https://jsfiddle.net/od1fL1kg/

jshjohnson:
@evu The latter is why it's not being merged. I'm currently working on getting decent test coverage so I can validate what breaks with this change.

evu:
@jshjohnson Just realised it's because the develop branch changed `setValueByChoice` to `setChoiceByValue` compared to the live branch.

tostringtheory:
@evu , @travist -

Please check out #310, a PR I just created that works to resolve this issue not by debouncing, as that introduces other issues with choices due to current code.  Specifically for us the issue is with callbacks that occur not firing with debounce.  310 utilizes a store/state for tracking a loading state, and if it is loading, then short circuits out of the render method.  What's nice is that when it sets that its not loading anymore, that also triggers render due to the existing subscribe.  Also, it's not having to clear/create a timeout repeatedly.  We saw a performance increase of minutes to seconds for 3700 items being rendered in.

If you guys could take a look and see if it works for you, or if its introduced other issues?

travist:
@tostringtheory I agree with your approach and approve. I have not had time to pull it in and test it myself, but I did give it a glance code review and it looks good. I would be just as happy with your fix.

jshjohnson:
Closed in favour of https://github.com/jshjohnson/Choices/pull/310

PR #306, opened 2017-12-08
Remove inline styling

### LINKED ISSUE #279: Remove inline styling from js to accommodate standard CSP policies
This line of code violates standard CSP policies due to the inline style being set:
```js
this.passedElement.setAttribute('style', 'display:none;');
```

What developers can do instead is specify the `hiddenState` class with one that has a `display: none;` rule when creating a Choices input. As there is an operation that occurs a few lines before:
```js
// Hide passed input
this.passedElement.classList.add(
   this.config.classNames.input,
   this.config.classNames.hiddenState,
);
```

Is there a case that requires this styling to be set here that I'm failing to consider?

---
* [Link to code - choices.js:2729](https://github.com/jshjohnson/Choices/blob/master/assets/scripts/src/choices.js#L2729)
* [Content Security Policy - Unsafe inline styles](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#Unsafe_inline_styles)

AndyRN:
## What does this do?
* Removes `setAttribute('style', 'display:none;');` lines from the JS to prevent CSP violations.
* Adds the `display: none;` rule to the default `is-hidden` class.

## Relevant issues
* Fixes #279


src/styles/css/choices.css
```diff
@@ -356,4 +356,9 @@
   opacity: .5;
 }
 
+.choices[data-type*="select-one"] .choices__input.is-hidden,
+.choices[data-type*="select-multiple"] .choices__input.is-hidden {
+  display: none;
```
jshjohnson: I think `.is-hidden` should be set here as a util: https://github.com/AndyRN/Choices/blob/ab9a471517015598e65ed4ebe7e5ac874dbd3cf9/src/styles/scss/base.scss#L123

AndyRN: Happy to include the `is-hidden` class within the base style sheet if you'd prefer?

Just to note, it still needs to be set here as well to reach the required level of specificity. As you can see here, it is being outranked by the more specific style set later on.

image

Which is why, in this case for the multi-select, the full `.choices[data-type*="select-multiple"] .choices__input.is-hidden` selector needs to be specified.

jshjohnson:
You have some conflicts :)

AndyRN:
I've rebased this pull request to avoid having multiple merges in the commit history when this branch goes in.

Let me know if there's any other issues!

jshjohnson:
Cool, thanks 🎉

PR #310, opened 2018-01-03
PR #299 / #236 Alternate Fix | 'Loading' State Tracking

tostringtheory:
This change was done off of a fork of master.  I understand that mainline new `dev` is done in develop, but I felt this was an urgent enough matter due to performance with the plugin, and a minor amount of change that it may be allowed.  This allows bulk updates to occur without re-rendering the list in O(n^2) efficiency in rendering.

The change resulted in a total of 67 lines changed in src files.  Created a new reducer to track a 'general' state for choices metadata state, such as if it is loading or not.  Modified the store to return it's loading state, and then wrapped any sections in choices.js that manipulate the groups/choices with a call to `_setLoading`, and made `render` check the loading status to so that it wouldn't attempt to render while it was doing non-interactive functions.  Due to the nature of how this fix is implemented, the rest of the calls seem to be working correctly, and don't fail such as the methods attempting to solve this issue with debounce, which was our first attempted solution until we saw the issues with callbacks.

The change in performance for our list (3700 items, rendering 10 with typeahead search) went from MINUTES to < 1 second.

Please do consider this for mainline to master to assist with people that have already been facing these issues as indicated by the PR's and issues.  I can't dedicate more time for the next couple weeks to PR into develop due to travel, but let me know if you would like me to do that.

Thanks for a great plugin!


tostringtheory:
@jshjohnson - I'm not quite sure what happened with Travis for this build, it looks like the service may have been having an outage, as it couldn't reach Github a couple times and failed at a bunch of points (not appearing to be build/code related), and then timedout with NPM too.  Are you able to requeue the build or take a look at it, or is there something you need me to do?

travist:
👍

jshjohnson:
Thanks for spending the time looking into this - I’ll review this ASAP :)

tostringtheory:
@jshjohnson - no worries.  We definitely like choices, but when we changed from a couple hundred items in a list, to ~4000, it became unusable.  Thankfully the fix wasn't that bad.

tostringtheory:
@jshjohnson -  

Woohoo!  Thanks for accepting!  Done dev for quite some time, but that was my first PR.  I may look at some of the other issues when I get back in town and attempt to kill one or two of them.

Thanks!

jshjohnson:
No worries - great work 👌🏻

PR #312, opened 2018-01-08
Improvements

YuryRybakSoftensity:
Added:
 * possibility to specify customer item comparer
 * 'stopSearch' event

Also performed some refactoring.


assets/scripts/src/choices.js
```diff
@@ -1508,7 +1508,7 @@ class Choices {
     // Run callback if it is a function
     if (this.input === document.activeElement) {
       // Check that we have a value to search and the input was an alphanumeric character
-      if (value && value.length >= this.config.searchFloor) {
+      if (value.length >= this.config.searchFloor) {
```
travist: If value is ```null``` this will cause a JS error. Why did you remove the value check?

YuryRybakSoftensity: You should take a look a little bit above: there is check for empty `value` already.

jshjohnson:
Thanks for this - unfortunately the `develop` branch is the branch being used for future releases (master is quite outdated). Would you be able to re-open this with develop as the base branch please?

YuryRybakSoftensity:
@jshjohnson Please take a look @ https://github.com/jshjohnson/Choices/pull/314

PR #326, opened 2018-01-22
IE 10/11 dropdown opening fix - it opened to the top instead to the bottom...

res0:
# The problem:
This bug shows just on the IE10/11 browser - when you go on the very end of the Choices demo page and click on the Cities select, then the dropdown opens to the bottom instead to the top even though it doesn't have enough space in the bottom and it makes you scroll down out of the viewport, which is unfortunate. I've tested it on IE10, IE11, Edge, Chrome, Firefox and it appears to be just IE10/11 bug.

# This is my solution:
**I've updated shouldFlip method to return true if dropdownPos parameter is not passed.** (was returning false before and from some reason, IE10 doesn't pass dropdownPos when the dropdown should flip up)
I've updated also the container test for shouldFlip function as it was expecting to return false if dropdownPos was not passed.
After this small update, the flip now works even on the IE10.


src/scripts/src/components/container.js
```diff
@@ -59,7 +59,7 @@ export default class Container {
    */
   shouldFlip(dropdownPos, windowHeight = getWindowHeight()) {
     if (dropdownPos === undefined) {
-      return false;
+      return true;
```
jshjohnson: Does this actually solve the problem? It seems this change would cause the dropdown to always flip irrespective of if needs to? 🤔

res0: Yes, it solved the problem and doesn't cause the dropdown to always flip (as far as I tested), I'll attach the demonstration video here:
[choices-ie10.zip](https://github.com/jshjohnson/Choices/files/1689150/choices-ie10.zip)

jshjohnson:
Closing due to inactivity and amount of conflicts

PR #329, opened 2018-01-25
Bind event listeners to component container instead document

basgren:
When Choices is integrated inside other js-components that consume events, they don't bubble to `document` and Choices stops working. Binding event listeners to Choices container instead `document` solves the issues.


jshjohnson:
This will be changing as part of the next release anyway so will not be necessary. Thanks anyway 👍

basgren:
Glad to hear that! Thanks!

PR #331, opened 2018-02-01
Add ability to configure the item comparer function.

travist:
We need the ability to configure the item comparer using something like [Lodash.isEqual](https://lodash.com/docs/4.17.4#isEqual). With this, we can do the following...

```js
import _isEqual from 'lodash/isEqual';
...
...
let choices = new Choices(...., {
  itemComparer: _isEqual
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,9 @@ Or include Choices directly:
     maxItemText: (maxItemCount) => {
       return `Only ${maxItemCount} values can be added.`;
     },
+    itemComparer: (choice, item) => {
+      return choice === item;
+    },
     classNames: {
       containerOuter: 'choices',
       containerInner: 'choices__inner',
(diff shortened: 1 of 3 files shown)


jshjohnson:
Nice @travist - thanks 💯

travist:
Thanks for merging this @jshjohnson. What is the possibility to get this back-ported into the 3.x branch? We will make a pull request against master if that makes it easier for you.

Right now we are [duck-punching](https://github.com/formio/formio.js/blob/master/src/components/select/Select.js#L16) a Choices method within [formio.js](https://github.com/formio/formio.js) to allow us to have deep equality checks. Having this in the 3.x choices would save us from that hack.

jshjohnson:
@travist  Sure, feel free!

travist:
Here it is thanks. https://github.com/jshjohnson/Choices/pull/334

PR #333, opened 2018-02-02
Add custom properties to 'addItem' event

BJJLangedijk:
## This is the problem:

Previously users were unable to get the customProperties set when adding an item with custom properties.

## Steps to reproduce:

- build item Object with custom properties.
- Add `addItem` eventlistener.
- create new Choices() instance and pass along item with customer properties.
- `addItem` eventlistener will be triggered however it will not contain custom properties.

## This is my solution:

Add customProperties values to the `triggerEvent(this.passedElement, 'addItem', params)`


jshjohnson:
Thanks. Would you mind also updating the README for this event please 👍

sandric:
I think it would be actually neat to describe it somewhere in docs. To find info how to programmatically add new option (choice) to select I had to search in issues.

PR #339, opened 2018-02-15
Replace hardcoded color for input with $choices-bg-color variable

waltton:


src/styles/scss/choices.scss
```diff
@@ -276,7 +276,7 @@ $choices-icon-cross-inverse: url(data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMjEiI
 .#{$choices-selector}__input {
   display: inline-block;
   vertical-align: baseline;
-  background-color: mix(#000000, #FFFFFF, 2.5%);
+  background-color: $choices-bg-color;
```
jshjohnson: Does this equate to the same colour?

waltton: Yes, `mix(#000000, #FFFFFF, 2.5%)` is equivalent to `#f9f9f9` that is the default value of `$choices-bg-color`.

jshjohnson: Cool, thanks 👌

PR #378, opened 2018-03-31
Fixed an issue where non-strings can be sent to the stripHTML method.

travist:


travist:
This should fix https://github.com/jshjohnson/Choices/issues/377

jshjohnson:
I'm a bit confused, how can `input.value` or `input.placeholder` be an object? It should always be a string

travist:
Here is a JSFiddle that shows the error. Try to select an item and look at the console when you do that. What you will find is that for any choices where a JSON is provided as the list of available options it crashes.

https://jsfiddle.net/travistidwell/1actk0fa/

This is a very common use case that we use within the Form.io core renderer @ https://github.com/formio/formio.js. Please verify on your end. Until a new release is out, we will need to duck-punch this method, which I don't really like doing.

Thanks,

Travis.

fernandoagarcia:
@travist if you are trying to save complex data types you can use customProperties to add additional data that can be accessed via the event.detail. I don't believe that using objects in values or placeholders is intuitive otherwise that would be a native element in HTML inputs. Integers I can understand but this seems like a very particular use case.  However, customProperties is the perfect key to use in your case: https://github.com/jshjohnson/Choices#items

travist:
I see your point, but one of the main motivations to using this library instead of HTML select input was its ability to manage complex data inputs as selectable values. The ONLY thing that is keeping this library from supporting this is the recent change to determine the input width, which should be determined via the "label" and not the "value". I may work on this pull request to determine if I have access to the label of the element instead of the value when determining the width and then it should work just fine.

My issue with using the ```customProperties``` is that there is no way of determining what the "value" should be that represents the object that is being stored. We introduced the ```itemComparer``` capability in a previous commit that would allow you to determine value equality using ```_.isEqual```, but now this will be very difficult since all items are identified by their value and customProperties has no meaning to the value of the item. Do we now create object Hashes as the values?... hard to say, and adds an enormous amount of overhead to something that worked very well the way it was before. Also, looking at the example you gave me, this now needs to be rewritten since technically this library cannot handle an ```array of objects``` anymore. 

I really wish the maintainer will re-consider keeping the ability to select complex data objects. Like I mentioned before, this was working perfectly before, and we have a lot of people using our Form.io renderer that uses this library extensively. We will need to fork this repo to move forward and I really don't want to do that.

jshjohnson:
I'm not necessarily opposed to fixing your issue - I do think it's a strange use case though.
 
I do think the fix needs revisiting though. I think a cleaner solution would be to just cast `value` to a string within getWidthOfInput before running stripHTML. Then replace will always be defined. Alternatively you could alter stripHTML to just return the value if it's not a string.

travist:
I agree that what I provided needs work. I am going to investigate to see if I have access to the "label" in this method since that is something that is always a string. Give me some time to investigate.

jshjohnson:
Thanks @travist 🎉

YuryRybakSoftensity:
@jshjohnson @travist I have investigated this item and found, that if we change `choice.value` to `choice.label` [here](https://github.com/jshjohnson/Choices/blob/394bde313d0f8a50b4c2b9d64d35b76ded68a515/assets/scripts/src/choices.js#L1264) it will resolve the issue.

YuryRybakSoftensity:
Here is the fix https://github.com/jshjohnson/Choices/pull/392

travist:
@YuryRybakSoftensity 👍

jshjohnson:
This has been resolved in develop - a value is now always cast to a string: https://github.com/jshjohnson/Choices/blob/develop/src/scripts/components/input.js#L24

travist:
I would like to express my concern with now making this library only support "string" values. The ability for this library to work with string and object values was the main purpose we chose this library as the basis for the select elements within the Form.io renderer. If this library moves in that direction, then we will either need to fork this repository or find another one, which I would really not like doing.

The proposed fix above does actually fix the problem, since it seems that the only place that this library currently breaks when the values are not strings was the placeholder text. Your statement on that fix where you state that it reflects a bigger problem where this library assumes the values are "strings" and that this library should force values to be strings is incorrect IMO. This library should be able to support objects as values like any other "select" library out there like https://selectize.github.io/selectize.js/ and https://select2.org/.

Please reconsider this support.

travist:
You mentioned in a previous comment that you feel it is a strange use-case to have a select dropdown select complex objects, so I wanted to provide some use cases in our renderer where this is needed.

Consider this CodePen.  https://codepen.io/travist/pen/XYrqpE

The submission data of the form is actually represented below the form, and this is how it is submitted to our backend. We have this type of form all over in our forms, where the selection of an item represents the selection of a more complex data item that is stored with the submission. This is the ONLY way to get "references" selected within our resources or API endpoints to form integrations between multiple submissions and complex data objects.

Your latest version of Choices.js completely breaks this for us and we cannot use it.

jshjohnson:
@travist thanks for the info, I'll look into it tomorrow to see if there's a clean solution 👍

michaelpumo:
I solved this issue for myself by casting **all values** in my options object, to strings. Choices.js appears to expect everything in an options object top be a string.

I had initially had **true numbers** as my **values** and it was causing the error above when clicking each of them.

e.g.

```
const myOptions = [
  { value: 1, label: 'Gold' },
  { value: 2, label: 'Silver' },
  { value: 3, label: 'Bronze' }
]

const newOptions = myOptions.map(item => ({ value: item.value.toString(), label: item.label.toString() }))
```

This will result in:

```
[
  { value: '1', label: 'Gold' },
  { value: '2', label: 'Silver' },
  { value: '3', label: 'Bronze' }
]
```

...which is Choices.js safe.

Hope this helps somebody else out there! 

If you still need values that are selected as numbers, you can always cast them back with something like `parseInt(value, 10)`

PR #384, opened 2018-04-18
Fix xss vulnerability(escape html in item label)

MaksimShakavin:
The vulnerability can be still reproduced, when user past the xss payload(`<svg onload=confirm(document.cookie)> & `) to the input field and add it as an item.


jshjohnson:
Closing due to master branch being frozen (this is already in develop) 👍

Issue #377, opened 2018-03-31
Uncaught TypeError: e.replace is not a function


travist:
There is an issue with the latest 3.0.4 branch that was caused by this commit.

https://github.com/jshjohnson/Choices/commit/394bde313d0f8a50b4c2b9d64d35b76ded68a515

The issue is that when I have select items that are objects, this fails with the following error.

```
Uncaught TypeError: e.replace is not a function
    at t.stripHTML (formio.full.js:16252)
    at Object.addItemText (formio.full.js:16250)
    at e.value (formio.full.js:16250)
    at e.value (formio.full.js:16250)
    at e.value (formio.full.js:16250)
```

It seems pretty clear that this was written for the select options that are strings, but did not take into account the select options that are objects where ```replace``` is not a function.

travist:
Fixed by https://github.com/jshjohnson/Choices/pull/378

travist:
Here is a JSFiddle with the issue.

https://jsfiddle.net/travistidwell/1actk0fa/

jshjohnson:
I think a cleaner solution would be to just cast value to a string within `getWidthOfInput` before running stripHTML. Then replace will always be defined.

travist:
I added a comment to the pull request that I would appreciate you reading and providing your input. One thing that I remembered when configuring this element is that you do allow for a "label" and a "value". Why are we not determining the width based on the "label" and not the "value". Maybe that would be cleaner since in every case the "label" will always be a string. That is something that is consistent no matter the type of value.

jshjohnson:
getWidthOfInput is called when you are inputting data into the input field - this wouldn't be a label, it would be the actual value of the input 👍

andershagbard:
A dirty hotfix would be to add a prototype function for Number and Object types.

`
Number.prototype.replace = function() {
    return this;
};

Object.prototype.replace = function() {
  return this;
};
`

MeekLogic:
Hopefully, this issue can be resolved soon...

jshjohnson:
This has been fixed in version [6.0.2](https://github.com/jshjohnson/Choices/releases/tag/v6.0.2)

PR #392, opened 2018-05-16
Fixed issue #377.

YuryRybakSoftensity:


assets/scripts/src/choices.js
```diff
@@ -1261,7 +1261,7 @@ class Choices {
     });
 
     if (choice && !choice.selected && !choice.disabled) {
-      const canAddItem = this._canAddItem(activeItems, choice.value);
+      const canAddItem = this._canAddItem(activeItems, choice.label);
```
jshjohnson: This doesn't make sense to me. `_canAddItem` expects the value of the choice to determine whether it can be added as an item 🤔

YuryRybakSoftensity: At all cases, which I tried all is working fine for me and it resolves #377 when value might be not string value. Label is always a string and in test cases, which I added to this fix you can see, that I'm searching choice by the label and it's added successfully, but if I leave `choice.value` it'll throw an error.
Could you provide me with case, where my changes break something?

jshjohnson: So if you're label and value are different values, this will cause an issue. This is more covering up a smell elsewhere - the actual problem is that the code is assuming the value is a string

jshjohnson:
Closing as a fix has been put into develop to ensure values are always strings 👍

assets/scripts/src/choices.js
travist: Please re-consider your position on this. I explained in further detail @ https://github.com/jshjohnson/Choices/pull/378#issuecomment-392547687

PR #396, opened 2018-05-24
TS project was breaking because of out dated TS definition 'itemOption' instead of 'itemChoice'

superelement:
TS project was breaking because of out dated TS definition 'itemOption' instead of 'itemChoice'. I've updated in README.md and TS def.

### DIFF
--- a/README.md
+++ b/README.md
@@ -484,7 +484,7 @@ classNames: {
   item: 'choices__item',
   itemSelectable: 'choices__item--selectable',
   itemDisabled: 'choices__item--disabled',
-  itemOption: 'choices__item--choice',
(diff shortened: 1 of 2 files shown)


superelement:
Not sure why the CI test is failing. I only changed the TS definition and README files. 
Btw, I did this on master branch, rather than develop (which I can see you've already fixed there), because it's kind of a critical error (for me anyway), as I need to change this CSS class name and TS won't compile without it. Would be great if you could incorporate it soon.

jshjohnson:
The master branch is frozen in its current state (I know there are issues with it). The next release will be from develop.

PR #398, opened 2018-06-06
add aria-live attribute to get screenreader to speak non-focused option

ryan-mahoney:
## This is the problem:
When using arrow keys to navigate through selections, screen reader does not read the highlighted item.

## Steps to reproduce:
On a Mac, enable Voiceover. Tab into a `Choices` component and observe that while the arrow keys are doing a great job of highlighting options, the screenreader is unaware that highlighted options have changed.

## This is my solution:
[ARIA Live Regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) I have added the additional HTML attribute `aria-live` to mirror the `aria-selected` behavior, toggling it between `off` and `assertive`.  In my testing, the screen reader is now reading the options as I navigate through them with the keyboard. Not sure how to create a unit test for this.


ryan-mahoney:
I'm going to close this PR and send something different as I'm testing and need to make some additional changes.

PR #400, opened 2018-06-07
expose new event that triggers when choice is highlighted

ryan-mahoney:
## This is the problem:
When using arrow keys to navigate through selections, screen reader does not read the highlighted item.

## Steps to reproduce:
On a Mac, enable Voiceover. Tab into a `Choices` component and observe that while the arrow keys are doing a great job of highlighting options, the screenreader is unaware that highlighted options have changed.

## This is my solution:
- add a new event `highlightedChoice`
- listen for that event in my own integration to update an `aria-live` region in my page.

---
Then in my instance of `Choices.js` I can listen for that event and update an `aria-live` region of my page that causes the screenreader to read every choice.

JS Code:
```
function addAriaLiveNode (containerEl, text) {
  while (containerEl.firstChild) {
    containerEl.removeChild(containerEl.firstChild);
  }
  const speakEl = document.createElement("span");
  speakEl.innerHTML = text;
  containerEl.appendChild(speakEl);
}

const speakerEl = [...document.querySelectorAll("[data-type='speaker']")].shift();

const selectors = [
  ...document.querySelectorAll("select[data-type='choices']")
];

selectors.forEach(el => {
  el.addEventListener("highlightChoice", function(event) {
    const label = event.detail.el.innerHTML.trim();
    addAriaLiveNode(speakerEl, label);
  });
  const choices = new Choices(el, {});
});
```

Relevant HTML Code:
```
<div class="sr-only" data-type="speaker" role="region" aria-live="assertive"></div>
```
---

About Aria Live Regions:
[ARIA Live Regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions)


jshjohnson:
@ryan-mahoney thanks 🎉 Would you be able to update the README with this new event please?

ryan-mahoney:
Sorry, will update this today!

jshjohnson:
👌 🎉 thanks

PR #409, opened 2018-07-03
Update demo URL.

annapowellsmith:
This is the URL used at the live demo. This change is needed for IE11 support.


annapowellsmith:
Assume the build error isn't related to this change :)

PR #420, opened 2018-07-19
Bower no longer recommended

JoshuaDoshua:
This is the problem:
Bower is no longer recommended (by Bower maintainers)

This is my solution:

- remove `bower.json`
- add `keywords` to `package.json`
- replace bower with yarn in readme

Also, as of NPM v5 the `--save` flag is no longer necessary as it saves to dependencies by default


jshjohnson:
👍 thanks

PR #430, opened 2018-08-16
Fix for loosing focus while selecting the same choice

MaksimShakavin:
Dirty fix for the issue https://github.com/jshjohnson/Choices/issues/429


jshjohnson:
Closing as this was raised against the previous major version 👍

PR #434, opened 2018-08-23
Make it work in Internet Explorer

barnslig:
## This is the problem:
The development branch does not work in Internet Explorer. There are visual problems like non-hidden selects and functional problems like non-working multi-selects.

## Steps to reproduce:
Launch the development branch in IE.

## This is my solution:
* [IE does not support multiple-parameter `classList.add()` / `classList.remove()`](https://caniuse.com/#feat=classlist); Replace it with single-parameter calls
+ IE needs polyfills for `Array.prototype.find` / `Array.prototype.includes`; Include them from core-js


barnslig:
Eventually it makes sense to investigate how to use [`useBuiltIns: 'usage'`](https://babeljs.io/docs/en/next/babel-preset-env.html#usebuiltins) regarding the Polyfills.

jshjohnson:
Which version of IE? Choices only supports >IE11

barnslig:
The problem appears in IE 11 which happens to be the latest version of Internet Explorer.

czapkowicz:
Yeap, on IE11 this lib is pretty much useless.

stof:
instead of bundling polyfills, it might be better to document the polyfills needed for IE 11. This avoids shipping the same polyfills multiple time in a project due to third-party libraries shipping the polyfills by themselves.

PR #469, opened 2018-11-24
Add the "files" section to package.json

limonte:
No need for publishing images, HTML files and other useless files in npm dist. After merging this PR only these files will remain:

The `public/assest/images` folder will be also excluded from npm package, that's a bad idea to have all those useless for end-users kilobytes in npm.

Read more: https://docs.npmjs.com/files/package.json#files

---

I wish "tech leads" in the JS community would care about the published bundles :)

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,12 @@
   },
   "author": "Josh Johnson",
   "license": "MIT",
+  "files": [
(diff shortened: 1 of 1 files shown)


jshjohnson:
👍 I do care

PR #474, opened 2018-11-28
Fixes jshjohnson/Choices/#447

Nettsentrisk:
## This is the problem:
Current implementation forces invalid HTML by requiring use of placeholder attribute on an option element to make it act as placeholder.

## Steps to reproduce:
Run HTML validation of current recommended manner of implementing single select box.

## This is my solution:
Change placeholderOption so that it uses the option element that has an empty value attribute as the placeholder, as [per W3C recommendation](https://www.w3.org/TR/html52/sec-forms.html#the-select-element):

> If a select element has a required attribute specified, does not have a multiple attribute specified, and has a display size of 1; and if the value of the first option element in the select element’s list of options (if any) is the empty string, and that option element’s parent node is the select element (and not an optgroup element), then that option is the select element’s placeholder label option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,11 +385,11 @@ const example = new Choices(element, {
 
 **Usage:** Whether the input should show a placeholder. Used in conjunction with `placeholderValue`. If `placeholder` is set to true and no value is passed to `placeholderValue`, the passed input's placeholder attribute will be used as the  placeholder value.
 
-**Note:** For single select boxes, the recommended way of adding a placeholder is as follows:
+**Note:** For single select boxes, the placeholder will be the option element with an empty value, as per W3C recommendation:
 
 ```html
 <select>
-  <option placeholder>This is a placeholder</option>
+  <option value="">This is a placeholder</option>
   <option>...</option>
   <option>...</option>
   <option>...</option>
--- a/src/scripts/components/wrapped-select.js
+++ b/src/scripts/components/wrapped-select.js
@@ -7,7 +7,7 @@ export default class WrappedSelect extends WrappedElement {
   }
 
   get placeholderOption() {
-    return this.element.querySelector('option[placeholder]');
+    return this.element.querySelector('option[value=""]');
   }
 
   get optionGroups() {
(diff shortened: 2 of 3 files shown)


stof:
Support for `placeholder` should be kept for backward compatibility though

Nettsentrisk:
Only if this were to be implemented in a patch/minor version, I guess. If this is seen as a breaking change, then it should be rolled into a new major version. 

But since this per now is dependent on invalid HTML, it seems like backward compatibility only supports bad HTML practices, which should be avoided.

stof:
Well, it is not backward compatible, as it forces projects using this feature to change their HTML for things to keep working. That's the definition of backward compatibility.

I'm not saying we should not add support for the proper way based on `value=""`. But the JS code should still fallback to this non-standard `placeholder` attribute if there is no `value=""` element (and potentially log a warning if there is an element using the non-standard placeholder to push people to migrate). The doc of course should document only the proper way.
Doing a major version just for that would actually mean that many projects may stop updating the library (due to not switching to the new major). A new minor version with the new feature and a warning when using the deprecated one is much easier for the upgrade path (and warns devs using the deprecated feature instead of making the placeholder disappear silently for them).

Note that I totally agree that using a non-standard attribute to detect the placeholder is not a good idea here. I'm only saying that the way to migrate away from it could be better.

Nettsentrisk:
I leave it up to the repo author to decide regarding versioning etc. 

If the change is wanted in a minor version update, then they should add a fallback solution to allow for the previous method to work, whereas in a new major version this should no longer work.

You could also solve this by adding support for a data-attribute (e.g. data-placeholder) to allow for the same thing, although I don't quite see a use case where you would need that since the selected/first option in the list would be used as a placeholder if you do not implement an actual placeholder value option as per the HTML spec.

stof:
@jshjohnson why is stalebot configured to closed PRs if you don't even review them before ?

Nettsentrisk:
This is such a no-brainer fix, please! :)

jshjohnson:
@Nettsentrisk Would you mind incrementing a major version please? 👍

jshjohnson:
Closing this in favour of https://github.com/jshjohnson/Choices/pull/617 which keeps backwards compatibility 👍

Issue #483, opened 2018-12-11
regexFilter refinement


gkweb:
I have a use case to avoid duplicate keywords being added across multiple instances of choices. 
Does an easy solution exist for this already?

As a suggestion how about replace regexFilter with canAddItems callback. Or allow a function also. This will cover a broader range of use cases.

### Custom validation before adding item

```javascript
/**
 Only allow items longer than 3 chars and starting with the letter l only.
*/
callbackCanAddItems: function (val) {
  return (val.length > 3 && val[0] === 'l')
}
```

## Covers regex also
```javascript
/**
 This can cover regex also
 - Only allow Alphabetical between A-z
*/
callbackCanAddItems: function (val) {
  return (/[A-z]/.test(val))
}
```

PR #485, opened 2018-12-13
callback to filter items before adding

gkweb:
## Description

This adds a callback to filter the user adding an item until said condition is met. In the test updates this would be - A text instance will only allow the value `test`. This can be expanded on greatly beyond the limits of what a regex filter only offers.

## Motivation and Context

Solves #483
I found a use case in a project I'm working on where 2 choices instances needed to avoid duplicate keywords items between them. This cannot be solved with regexFilter only feature available right now as the user inputs are dynamic and change on the fly. 

I understand it's probably possible to write an overly complex regex pattern that updates with the `addItem` and `removeItem` event - But this is way more effort than the solution on offer in this PR.

Another use case this would cover is
`regexFilter` covers email validity and `callbackFilterAddItem` checks other values within form to pass this value.

## How Has This Been Tested?
I've copied the current structure and added tests in cypress

Tests passing locally before creating this PR

This will affect Text inputs only.

This is largely a non-breaking change as it's an addition to existing functionality.

Try it yourself like so: 

```javascript
new Choices('#choices-add-item-callback', {
  callbackFilterAddItem: function (value) {
    return (value !== 'test')
  }
});
```

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ will be returned. If you target one element, that instance will be returned.
     position: 'auto',
     resetScrollPosition: true,
     regexFilter: null,
-    callbackFilterAddItem: null,
+    addItemFilter: null,
     shouldSort: true,
     shouldSortItems: false,
     sortFn: () => {...},
@@ -340,7 +340,7 @@ Pass an array of objects:
 
 **Usage:** Whether the scroll position should reset after adding an item.
 
-### callbackFilterAddItem
+### addItemFilter
 **Type:** `Function` **Default:** `null`
 
 **Input types affected:** `text`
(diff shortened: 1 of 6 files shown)


gkweb:
@jshjohnson I haven't had a reply about this yet. Is it safe to assume you are not interested in this feature?

Cheers

jshjohnson:
@gkweb Thanks for this and sorry for the delay. Would you mind updating the read me with this new functionality please?

gkweb:
@jshjohnson - Readme Updated

README.md
```diff
@@ -92,6 +92,7 @@ will be returned. If you target one element, that instance will be returned.
     position: 'auto',
     resetScrollPosition: true,
     regexFilter: null,
+    callbackFilterAddItem: null,
```
jshjohnson: Thinking about it - I think this should just be named `addItemFilter` to align with `regexFilter` 🤔 What do you think?

gkweb: @jshjohnson - Agree. I'll update

jshjohnson:
Thanks 🎉

PR #496, opened 2019-01-18
Fixes issue where deepmerge concatenates arrays inside DEFAULT_CONFIG

jhou:
... instead of replacing them.

## Description
provides `deepmerge` with a function to replace its typical array merging behavior with a function that overwrites the destination array with a copy of the source array instead of concatenating the two.
Documentation for that feature here: https://www.npmjs.com/package/deepmerge#arraymerge

## Motivation and Context
Without this fix, the `searchFields` config will *always* have `[ 'label', 'value']` plus whatever config the user passes in, making it impossible to NOT search both `label` and `value`
https://github.com/jshjohnson/Choices/issues/495

## How Has This Been Tested?
Tests included in PR.    Basically, attempt to type `value` entries into the search input, and make sure there are no results.

## Types of changes

## Checklist:

- [x] My code follows the code style of this project. (i hope)

### DIFF
--- a/cypress/integration/select-multiple.spec.js
+++ b/cypress/integration/select-multiple.spec.js
@@ -853,5 +853,35 @@ describe('Choices - select multiple', () => {
           });
       });
     });
+
+    describe('searching by label only', () => {
+      it('gets zero results when searching by value', () => {
+        cy.get('[data-test-hook=search-by-label]')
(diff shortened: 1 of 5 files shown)


jshjohnson:
Thanks for this :)

PR #517, opened 2019-02-13
Fix #499 documentation error about passedElement

### LINKED ISSUE #499: Documentation error
It seems there is a bug with your docs: 

```
// or
const example = new Choices(document.getElementById('example'));

example.passedElement.addEventListener('addItem', function(event) {
  // do something creative here...
  console.log(event.detail.id);
  console.log(event.detail.value);
  console.log(event.detail.label);
  console.log(event.detail.customProperties);
  console.log(event.detail.groupValue);
}, false);
```

That doesn't work. After some debugging, I found it should be:

`example.passedElement.element.addEventListener('addItem', function(event) {`

Spone:
see #499

### DIFF
--- a/README.md
+++ b/README.md
@@ -589,7 +589,7 @@ element.addEventListener('addItem', function(event) {
 // or
 const example = new Choices(document.getElementById('example'));
 
-example.passedElement.addEventListener('addItem', function(event) {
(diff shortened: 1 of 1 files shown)


jshjohnson:
Thanks 👍

PR #536, opened 2019-02-28
Remove the CustomEvent polyfill from the bundle

stof:
## Motivation and Context
https://github.com/jshjohnson/Choices/issues/519#issuecomment-465271930 says the CustomEvent polyfill was kept in the library because it cannot be loaded from polyfill.io. but it actually can be loaded.

## Types of changes
This could be considered as a BC break for people supporting old browsers, as they now need to load an additional polyfill (the polyfill.io URL in the readme got updated).

## Checklist:


jshjohnson:
Thanks 🎉

axten:
imho a bad decision because core-js do not include a CustomEvent Polyfill.
So auto polyfilling with babel dont work anymore.

stof:
@axten but that's also true for `Element.prototype.classList` or `window.requestAnimationFrame`, as core-js does not include web polyfills (yet) but ES polyfills.
The rules set by @jshjohnson previously is to bundle polyfills only if they cannot be loaded easily from polyfill.io
Forcing to bundle the polyfill would lead to loading multiple polyfills for CustomEvent if any other library also loads one.

axten:
hmpf, actually all browsers in my scope have classList and requestAnimationFrame, so CustomEvent is the only problem for me now... we dont want to use polyfill.io so I now have to add a polyfill manually or stick to version 6

stof:
@axten polyfilling yourselves is still a better choice. If 2 libraries you use need to use CustomEvent, you don't want both of them to load the polyfill, as the browser would end up downloading it twice.

Issue #503, opened 2019-01-25
IE11 - Scroll bar closes dropdown immediately on click if select is single-select with text input


markjoeljimenez:
If a user decides to click on the scroll bar to navigate through the list on IE11, the dropdown closes immediately.

This issue is happening on your demo site as well: https://joshuajohnson.co.uk/Choices/

DarrenMack-OD:
I have the same issue on my site as well :( It is frustrating as most of my clients are corporate on IE11 :/

markjoeljimenez:
As a temporary solution, we're simply hiding the scroll bar on IE.

```
.choices[data-type*='select-one'] .choices__list--dropdown .choices-list {
    -ms-overflow-style: none;
}
```

DarrenMack-OD:
Thanks @markjoeljimenez We might give that a go ;)

DarrenMack-OD:
@markjoeljimenez Weird, that CSS doesn't work for me :/

PR #547, opened 2019-03-12
Fix bug ScrollBars in IE11

DarrenMack-OD:
## Description
Fixes scrollbar issue in IE11 (#503). There is an EDGE-CASE where-by if a user were to select the same option twice, the dropdown doesn't go away. This is, however a very extreme case and we are on the right track

## Motivation and Context
No one else seemed to have tried yet, so I thought I would give it a go

## How Has This Been Tested?
Tested in all modern browsers and on a Windows7 VM with IE11

## Types of changes

## Checklist:

### DIFF
--- a/public/test/select-multiple.html
+++ b/public/test/select-multiple.html
@@ -16,12 +16,12 @@
   <meta name="theme-color" content="#ffffff">
 
   <!-- Ignore these -->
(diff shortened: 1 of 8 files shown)


DarrenMack-OD:
If it's ok, could we merge this soon please. I have a fix for the webPack problem I would like to try as well, but it keeps adding it to this pull-request :(

jshjohnson:
Thanks 🎉

PR #566, opened 2019-04-02
Fix CustomEvent creation in IE9+

zackschuster:
## Description
the [constructor API](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent) isn't supported in IE, requiring a polyfill for those environments.

## Motivation and Context
since a simple solution exists -- use the (now-deprecated) [stateful construction API](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/initCustomEvent) in IE -- and since CustomEvent creation is highly localized, it seemed trivial to implement.

with this change, modern browsers will use the recommended form while IE will gracefully fall back to the deprecated form.

this should also remove the need to load a polyfill for those browsers.

## How Has This Been Tested?
using CI. it only adds a `try`/`catch` block, so the original code flow is undisturbed.

## Types of changes

## Checklist:

### DIFF
--- a/src/scripts/lib/utils.js
+++ b/src/scripts/lib/utils.js
@@ -178,11 +178,17 @@ export const sortByAlpha = (a, b) => {
 export const sortByScore = (a, b) => a.score - b.score;
 
 export const dispatchEvent = (element, type, customArgs = null) => {
-  const event = new CustomEvent(type, {
-    detail: customArgs,
-    bubbles: true,
-    cancelable: true,
-  });
+  let event;
+  try {
+    event = new CustomEvent(type, {
+      detail: customArgs,
+      bubbles: true,
+      cancelable: true,
+    });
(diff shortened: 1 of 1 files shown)


jshjohnson:
The last release intentionally removed the CustomEvent polyfill - you should instead bring your own polyfill: https://github.com/jshjohnson/Choices#browser-compatibility :)

zackschuster:
right -- i'm adding six lines of code so the vast majority of consumers won't need a polyfill.

jshjohnson:
But the vast majority of users don't use <IE11 - the aim of this library is to be as small as possible. Including code for unsupported browsers is unnecessary. The polyfill will be equally as small if you need to support older browsers

zackschuster:
okey doke. thanks for the replies 😄

PR #572, opened 2019-04-07
update type for item function in Template interface

krystofbe:
## Description
item function in Templace interface accepts three parameters. Current typescript typing only provides one parameter.

## Checklist:
Example in documentation needs to be updated, i added the parameter to the example function

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -116,7 +116,7 @@ declare namespace Choices {
     containerInner?: () => HTMLElement;
     itemList?: () => HTMLElement;
     placeholder?: (value: string) => HTMLElement;
-    item?: (data: any) => HTMLElement;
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -116,7 +116,7 @@ declare namespace Choices {
     containerInner?: () => HTMLElement;
     itemList?: () => HTMLElement;
     placeholder?: (value: string) => HTMLElement;
-    item?: (data: any) => HTMLElement;
+    item?: (globalClasses: any, data: any, removeItemButton: boolean) => HTMLElement;
```
jshjohnson: The majority of these look out of date from what I can see

duanecilliers: PR submitted for updating these types #619

jshjohnson:
I have merged #619 which includes this fix 👍 thanks

PR #574, opened 2019-04-09
Fix #573

### LINKED ISSUE #573: new -> destroy does not restore preset choices
I expect it restores passed input element to initial state, but was not.

https://github.com/jshjohnson/Choices#destroy
> Usage: Kills the instance of Choices, removes all event listeners and returns passed input to its initial state.

I use the version 7.0.0.
Here is a code: https://gist.github.com/kzkn/43ddb8a52712661c75dfea2d79edeb56

kzkn:
## Description

This PR fixes #573. It contains following fixes:

 - Restores preset choices after `destroy`
 - Loads preset choices only once for each choice after `init`

## How Has This Been Tested?

E2E tests are added.


cypress/integration/select-one.spec.js
```diff
@@ -861,5 +861,39 @@ describe('Choices - select one', () => {
           });
       });
     });
+
+    describe('new, destroy, init', () => {
```
jshjohnson:
```suggestion
    describe('re-initialising a choices instance', () => {
```

cypress/integration/select-one.spec.js
```diff
@@ -861,5 +861,39 @@ describe('Choices - select one', () => {
...
+      it('has 3 choices', () => {
+        cy.get('[data-test-hook=new-destroy-init]')
+          .find('.choices__list--dropdown .choices__list')
+          .children()
+          .should('have.length', 3)
+      });
+
+      it('has 3 choices after destroy', () => {
```
jshjohnson: This should have a describe around it with a description like `destroying the choices instance` and then the `it` description should be along the lines of `the original select element is preserved`

cypress/integration/select-one.spec.js
```diff
@@ -861,5 +861,39 @@ describe('Choices - select one', () => {
...
+          .find('button.destroy')
+          .click()
+
+        cy.get('[data-test-hook=new-destroy-init]')
+          .find('select')
+          .children()
+          .should('have.length', 3)
+      });
+
+      it('has 3 choices after destroy and init', () => {
```
jshjohnson: See comment above

public/test/select-one.html
```diff
@@ -346,6 +357,14 @@ <h2>Select one inputs</h2>
         new Choices('#choices-set-choice-by-value').setChoiceByValue('Choice 2');
 
         new Choices('#choices-search-by-label', { searchFields: ['label'] });
+
+        const newDestroyInitChoices = new Choices('#choices-new-destroy-init')
+        document.querySelector('button.destroy').addEventListener('click', () => {
+          newDestroyInitChoices.destroy()
+        })
+        document.querySelector('button.init').addEventListener('click', () => {
+          newDestroyInitChoices.init()
```
jshjohnson:
```suggestion
          newDestroyInitChoices.init();
```

public/test/select-one.html
```diff
@@ -346,6 +357,14 @@ <h2>Select one inputs</h2>
         new Choices('#choices-set-choice-by-value').setChoiceByValue('Choice 2');
 
         new Choices('#choices-search-by-label', { searchFields: ['label'] });
+
+        const newDestroyInitChoices = new Choices('#choices-new-destroy-init')
+        document.querySelector('button.destroy').addEventListener('click', () => {
+          newDestroyInitChoices.destroy()
```
jshjohnson:
```suggestion
          newDestroyInitChoices.destroy();
```

public/test/select-one.html
```diff
@@ -346,6 +357,14 @@ <h2>Select one inputs</h2>
...
+        })
+        document.querySelector('button.init').addEventListener('click', () => {
+          newDestroyInitChoices.init()
+        })
```
jshjohnson:
```suggestion
        });
```

public/test/select-one.html
```diff
@@ -346,6 +357,14 @@ <h2>Select one inputs</h2>
         new Choices('#choices-set-choice-by-value').setChoiceByValue('Choice 2');
 
         new Choices('#choices-search-by-label', { searchFields: ['label'] });
+
+        const newDestroyInitChoices = new Choices('#choices-new-destroy-init')
+        document.querySelector('button.destroy').addEventListener('click', () => {
+          newDestroyInitChoices.destroy()
+        })
```
jshjohnson:
```suggestion
        });
```

src/scripts/choices.js
```diff
@@ -219,12 +219,13 @@ class Choices {
...
 
+    this.clearStore();
+
     if (this._isSelectElement) {
       this.passedElement.options = this._presetChoices;
     }
 
-    this.clearStore();
-
+    this._presetChoices = [].concat(this.config.choices);
```
jshjohnson: Is `this._presetChoices` modified? Not sure why this is needed

jshjohnson:
Thanks for this - i've left some comments 👍

perifer:
Tested this, and can confirm that it works in my app. Great job!

src/scripts/choices.js
kzkn: It was [here](https://github.com/jshjohnson/Choices/blob/0829899284f7a45a9d9ee9cc3902fb14cbfd1bf8/src/scripts/choices.js#L1905-L1917).
I need this change to prepare for the next `init` call. But this change is incomplete.
`this._presetChoices` contains options those are built from `this.config.choices`. They should not restore to `passedElement.options`.

I will fix it with next commit.

kzkn: I commited 79a4ae0 . It solve the above issue.
Coping `this.config.choices` is necessary. It avoids appending duplicated items on second `init`.

cypress/integration/select-one.spec.js
kzkn: Is a0626507edee0a8db4a2d5a16ffdf65df7ac6463 your expected one...?

franke-mc:
@jshjohnson: Maybe you can comment on the current status of this?

perifer:
I'd love to see this resolved. Anything I can do to help?

kzkn:
@jshjohnson re-review please?

cypress/integration/select-one.spec.js
```diff
@@ -861,5 +861,39 @@ describe('Choices - select one', () => {
...
+
+        cy.get('[data-test-hook=new-destroy-init]')
+          .find('.choices__list--dropdown .choices__list')
+          .children()
+          .should('have.length', 3);
+      });
+    });
+
+    describe('destroying the choices instance', () => {
+      it('the original select element is preserved', () => {
```
jshjohnson:
```suggestion
      it('preserves the original select element', () => {
```

cypress/integration/select-one.spec.js
```diff
@@ -861,5 +861,39 @@ describe('Choices - select one', () => {
           });
       });
     });
+
+    describe('re-initialising a choices instance', () => {
+      it('the original choices element is preserved', () => {
```
jshjohnson:
```suggestion
      it('preserves the choices list', () => {
```

public/test/select-one.html
```diff
@@ -346,6 +357,14 @@ <h2>Select one inputs</h2>
         new Choices('#choices-set-choice-by-value').setChoiceByValue('Choice 2');
 
         new Choices('#choices-search-by-label', { searchFields: ['label'] });
+
+        const newDestroyInitChoices = new Choices('#choices-new-destroy-init')
```
jshjohnson:
```suggestion
        const newDestroyInitChoices = new Choices('#choices-new-destroy-init');
```

kzkn:
@jshjohnson
I had fixed new indicated items from you and a conflict with master branch.

Then, I have gotten a new ambiguity.
Should the `Choices` selection be reproduced in the original DOM after calling `destroy`?

```html
<select>
<option selected>one</option>
<option>two</option>
<option>three</option>
</select>

<script>
const select = document.querySelector('select')
console.log(select.value) // => one
const choices = new Choices(select)
// select "two" on UI
choices.destroy()
console.log(select.value) // what?
</script>
```

How do you think it should work?

Thanks!

jshjohnson:
> Should the `Choices` selection be reproduced in the original DOM after calling `destroy`?

For a first pass I would say no. We can revisit this is people want to though

kzkn:
@jshjohnson 
> For a first pass I would say no. We can revisit this is people want to though.

Thanks! I got it.

> P.s. you have some conflicts!

I have fixed.

filip-jk:
Can't wait for this to be released!

src/scripts/choices.js
```diff
@@ -144,6 +144,7 @@ class Choices {
         classNames: this.config.classNames,
         template: data => this._templates.option(data),
       });
+      this._presetOptions = this.passedElement.options;
```
jshjohnson: Could this be moved above line 186 please? That's where all the "preset" variables are defined

kzkn: Thanks! I have fixed.

filip-jk:
@kzkn Thank you, Sir!

@jshjohnson Thank you, as well! Is there any chance to get this updated version on NPM? :)

jshjohnson:
@filip-jk this has been released with 9.0.1 🎉 thanks @kzkn

filip-jk:
@jshjohnson Ace, thanks!

filip-jk:
@jshjohnson Sorry for bugging you about this but I think there is an issue with the destroy() method. It doesn't seem to pass the selected value? 

I'm on 9.0.1 and let's say, the select I initialize Choices on has the selected value of "3". I destroy that Choices object and, unfortunately, the selected value on the <select> tag is not "3". It gets "reset" I believe.

jshjohnson:
That's expected @filip-jk https://github.com/jshjohnson/Choices/pull/574#issuecomment-553794506

filip-jk:
@jshjohnson I'm confused. I'm initializing Choices on a select that has a selected value of "5". Choices take that selected value and I can see it in the rendered Choices dropdown menu. Once I destroy it, the select gets totally reset so it's not really the states the Choices was initialized on, right?

If that's expected than the description of the method is wrong:

"Kills the instance of Choices, removes all event listeners and returns passed input to its initial state."

It's not the initial state if the selected option/value is different.

jshjohnson:
Yeah you're right, I would expect the originally selected option to persist. I'll investigate

tagliala:
Hi, is it possible to reopen this issue? I'm experiencing the same issue described by @filip-jk

The use case is in a stimulus.js + turbo application, I will try to provide a reduced test case, but I'm going to describe it

1. Open a page with a Choices.js input with a selected value of 3
2. Go back in history (that will call `destroy` on the Choices.js instance)
3. Navigate forward to the page of Choices.js (that will trigger a `init` on the Choices.js instance)
4. I expect `3` to be selected, but the selection is lost

Issue #473, opened 2018-11-28
Can't disable first item on initiate


markjoeljimenez:
Not sure if intended, but I can't seem to disable the first option.

https://jsfiddle.net/uf9bmpy6/

astr0sl0th:
I am having the same issues too:

![screenshot 2019-02-07 at 16 06 05](https://user-images.githubusercontent.com/9317246/52424850-bba33b80-2af2-11e9-9456-286b1c4a92f1.jpg)

PR #583, opened 2019-05-13
Fix first option if disabled (#473)

markjoeljimenez:
### Description
Fix issue [#473](https://github.com/jshjohnson/Choices/issues/473).
The first disabled choice passed in as an option is not actually disabled.

### Reproduced example
https://jsfiddle.net/uf9bmpy6/

### Fix
We can simply pass in the disabled choice instead of checking whether it's preselected or not.


jshjohnson:
Thanks for this @markjoeljimenez. Could you add some tests to cover this change please?

markjoeljimenez:
Hey @jshjohnson, I haven't written tests before so please let me know if my tests are sufficient enough.

cypress/integration/select-one.spec.js
```diff
@@ -861,5 +861,39 @@ describe('Choices - select one', () => {
...
     });
+
+    describe('disabled-first-by-options', () => {
+      beforeEach(() => {
+        cy.get('[data-test-hook=disabled-first-choice-by-options]')
+          .find('.choices')
+          .click();
+      });
+
+      it('first choice is disabled', () => {
```
jshjohnson:
```suggestion
      it('disables the first choice', () => {
```

cypress/integration/select-one.spec.js
```diff
@@ -861,5 +861,39 @@ describe('Choices - select one', () => {
...
+        cy.get('[data-test-hook=disabled-first-choice-by-options]')
+        .find('.choices__list--dropdown .choices__list')
+        .children()
+        .first()
+        .should('have.class', 'choices__item--disabled');
+      });
+
+      let selectValue;
+  
+      it('first non-disabled choice is selected', () => {
```
jshjohnson:
```suggestion
      it('selects the first non-disabled choice', () => {
```

public/test/select-one.html
```diff
@@ -230,6 +236,29 @@ <h2>Select one inputs</h2>
           removeItemButton: true,
         });
 
+        new Choices('#choices-disabled-first-choice-by-options', {
```
jshjohnson:
```suggestion
        new Choices('#choices-disabled-choice-via-options', {
```

jshjohnson:
Hey - i've opened a new PR with your fix + some tweaks over latest master. Thanks 👍

PR #259, opened 2017-09-21
Addressing issue #235

maximmig:
https://github.com/jshjohnson/Choices/issues/235: Rendering of selects with mixed options and optgroups in root level returns only groups - fixed.
Also corrected some typos.

Issue #235, opened 2017-08-23
Rendering of selects with mixed options and optgroups in root level returns only groups


Igloczek:
I tried to use your truly great lib to beautify selects in my Redmine tasks management site, but unfortunately, almost everywhere selects have in root level mixed `option` and `optgroup` tags, but this type of structure seems to not be supported here, b/c only groups are rendered.

Example
```html
<select>
  <option value="1">User</option>
  <option value="2">User</option>
  <option value="3">User</option>
  <optgroup label="Group">
    <option value="10">Users group 1</option>
    <option value="20">Users group 2</option>
    <option value="30">Users group 3</option>
  </optgroup>
  <option value="4">User</option>
  <option value="5">User</option>
</select>
```

I tried to fix this issue modifying `_createInput` and `render` methods but looks like there is a big bold line drawn between groups and options, which makes fixing this impossible without changing lots of rendering and parsing logic 😢 

I manage to render both types, but options from groups were also printed with root level options and the order was different than in source select code, b/c it render it by type.

IMO to fix it in a right way, all options and groups should be stored together in one object and then rendered together in proper order.

Do you have any other idea how to fix it or maybe there is some work in progress?

maximmig:
I can fix it after https://github.com/jshjohnson/Choices/pull/257 gets merged. I've managed to do it on my end.

EvanLovely:
This bug still is present on 6.0.3 & 7.0.0

PR #592, opened 2019-05-23
Rendering of selects with mixed options and optgroups in root level returns only groups

EvanLovely:
Addressing issue #235 
Trying to pick up where PR #259 left off (there were merge conflicts to resolve)


EvanLovely:
Nevermind; I think that due to v6 => v7 this PR is too out of date to rescue...

Issue #551, opened 2019-03-18
Typescript type info 'choices' missing


Moeriki:
To get up and running faster it'd be convenient to have type info on the `choices` option. Right now it is set to `any[]`.

I have written this for myself, based on the examples I saw in the docs. Not sure how correct it is.

```ts
export interface ChoiceItem {
  customProperties?: { [key: string]: string };
  disabled?: boolean;
  label: string;
  selected?: boolean;
  value: string;
}

export interface ChoiceGroup {
  choices?: ChoiceItem[];
  id?: string;
  label: string;
}

export type Choice = ChoiceItem | ChoiceGroup;
```

Can anyone confirm the validity?

Would a PR be accepted adding this to the project?

jshjohnson:
Sure :) sounds good

PR #609, opened 2019-07-03
Fix 551: Add type info for choice, group, and item

Moeriki:
Refs: #551

## Motivation and Context

PR was motivated in #551 

## Description

- TypeScript definition docs reference the latest published version of the project. And the latest TypeScript `^2`.
- All properties are optional except `label` and `value`.
- `value` is type `any`. I think users can attach anything to this?
- `customProperties` is typed as `{ [prop: string]: any }`. Not sure if `any` would be better.

Your thoughts?

## How Has This Been Tested?

Not.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 node_modules/
+types/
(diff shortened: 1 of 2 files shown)


jshjohnson:
Thanks 👍

PR #617, opened 2019-07-18
Add support for standard-compliant placeholder option

### LINKED ISSUE #447: placeholder attribute on option is invalid
For the config it says:
> Note: For single select boxes, the recommended way of adding a placeholder is as follows:
> 
> ```
> <select>
>   <option placeholder>This is a placeholder</option>
>   <option>...</option>
>   <option>...</option>
>   <option>...</option>
> </select>
> ```
Yet `placeholder` is not a valid attribute for the option element. 

The way this works in the HTML spec is that "placeholder" `option` is given an empty value attribute. From the [HTML spec](https://w3c.github.io/html/sec-forms.html#placeholder-label-option):
> If a select element has a required attribute specified, does not have a multiple attribute specified, and has a display size of 1; and if the value of the first option element in the select element’s list of options (if any) is the empty string, and that option element’s parent node is the select element (and not an optgroup element), then that option is the select element’s placeholder label option.

Choices should respect this way of marking up a `select`'s placeholder option with valid HTML.

stof:
## Description
Fixes #447

Replaces https://github.com/jshjohnson/Choices/pull/474. The implementation is almost the same. But my PR preserves backward compatibility by using `<option placeholder>` as a fallback when there is no `<option value="">`.

## Types of changes

## Checklist:


src/scripts/components/wrapped-select.js
```diff
@@ -7,7 +7,10 @@ export default class WrappedSelect extends WrappedElement {
   }
 
   get placeholderOption() {
-    return this.element.querySelector('option[placeholder]');
+    return (
+      this.element.querySelector('option[value=""]') ||
+      this.element.querySelector('option[placeholder]')
+    );
```
jshjohnson: Would you mind just adding a comment here explaining why both are supported please 👍

stof: @jshjohnson done.

jshjohnson: Would you also mind incrementing a minor version please (not major - ignore my original comment )

stof: done (as a separate commit, as updating the build file for the new version also included other changes which were not yet in it).

tinovyatkin: That should be done in the same `.querySelector('option[value=""], option[placeholder]')`

stof: @tinovyatkin no, that's not the same. Your proposal would use whatever is found first. My code is using the standard way if it is there (and so if you define both, the standard way wins)

tinovyatkin: @jshjohnson 
Ugh, if somebody defines both, one with invalid, but very implicit and documented (by this library) attribute `placeholder` and second one just with empty value and empty value starts winning - I think it's a breaking change.

PR #644, opened 2019-09-27
Use event delegation

### LINKED ISSUE #643: Use event delegation?
I noticed the library adds a new copy of each event listener to the document element for each instance of Choices on the page.

This could be made more performant by leveraging a simple event delegation system.

bronzehedwick:
## Description
Instead of attaching a new root-level event listener for bubbling events
for every choices instance, use a simple event delegation script to
handle each event type.

Each event callback function already is coded as if it were fully
delegated, since the events are attached at the document level, so
no changes are needed to detect which element is being called.

Note that focus and blur event do not bubble, so they have been left as
they are.

Also note that the event delegation uses an IIFE purposely instead of ES6 modules, since the event list should be globally cached, and it doesn't make sense to instantiate a new scope for each instance (then we're back where we started!)

## Motivation and Context
Using event delegation increase page speed, especially if there are many choices instances.

fix #643

## How Has This Been Tested?
I ran the standard tests via `npm test` on macOS 10.13.6.

I also tested the local server manually to make sure I could still perform all operations.

This change's scope is limited to adding an event delegation script and the `_addEventListeners` and `_removeEventListeners` functions.

## Types of changes

## Checklist:

### DIFF
--- a/src/scripts/choices.js
+++ b/src/scripts/choices.js
@@ -1,6 +1,7 @@
 import Fuse from 'fuse.js';
 import merge from 'deepmerge';
 
+import './lib/delegate-events';
 import Store from './store/store';
 import {
   Dropdown,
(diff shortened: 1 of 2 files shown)


bronzehedwick:
The tests are failing for this PR, but it's because cypress fails to start. I can't think of any code changes here that would cause that. And funny enough I'm seeing a similar issue with cypress failing to start for a work project. Is there some larger issue with cypress going on?

Here's the relevant log output.

<samp>
[22:15:14]  Verifying Cypress can run /home/travis/.cache/Cypress/3.1.5/Cypress [started]

[nodemon] 1.18.10

[nodemon] to restart at any time, enter `rs`

[nodemon] watching: *.*

[nodemon] starting `npm run css:build`

[22:15:15]  Verifying Cypress can run /home/travis/.cache/Cypress/3.1.5/Cypress [failed]

Cypress failed to start.
</samp>

jshjohnson:
If you merge in latest master, the CI build should be fixed 👍

guidefreitas:
This PR attenuate some of performance issues we are having when a select has a lot of options. 

Thanks!

bronzehedwick:
> If you merge in latest master, the CI build should be fixed 👍

I merged in master and the builds now pass. Thanks!

bronzehedwick:
@jshjohnson any chance this PR can get a review? Thanks!

jshjohnson:
Thanks @bronzehedwick 👍

markx3:
Awesome! :tada:

jshjohnson:
This fix is included in version 7.0.2 👍

PR #645, opened 2019-10-01
Fix search when option label has white space

dexcell:
## Description
Clean up whitespace in `label` in `haystack` before passing it to `fuse`

## Motivation and Context
Problem: Choicejs cannot search  very well when there is new line / white space inside the `<option>` html tag. Sometimes it will show no result even tough there is result .

Video:
choices-problem

This kind of option is very common when you create the html using php or other programming language because sometime we do `if else` or other logic inside the option tag which generate new line in html. e.g.
````
<option>
   Hello
                       (Description)
</option>
````

## Screenshots (if appropriate):
Console log haystack in `_searchChoices` function. Notice the label value before and after clean.
Screen Shot 2019-10-01 at 22 05 53
After
Screen Shot 2019-10-01 at 22 06 39

## Types of changes

## Checklist:


jshjohnson:
If you merge in latest master, the CI build should be fixed 👍

PR #648, opened 2019-10-02
fix unhighlightAll performance issue

Tomakava:
## Description
Due to an error in a condition, hasHighlightedItems variable is assigned an array and the condition below is always true. As a result, unhighlightAll is called each time, which updates items that need to be rendered again.

## Motivation and Context
When using Formio.js with large number of select components performance issue was noticed.
For every click and keypress in any component items of all select components are rendered.

## How Has This Been Tested?
Tested locally in Formio.js
After this fix unhighlightAll and _render are not called on every click

## Types of changes

## Checklist:

### DIFF
--- a/src/scripts/choices.js
+++ b/src/scripts/choices.js
@@ -1451,7 +1451,7 @@ class Choices {
         this.hideDropdown();
       }
     } else {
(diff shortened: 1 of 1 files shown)


jshjohnson:
Good spot! Thanks a lot 💯

PR #660, opened 2019-10-19
use templates from instance, fixes #461

### LINKED ISSUE #461: `option` is not calling from `callbackOnCreateTemplates` since 4.x
Hello,

Was `option(data)` removed in 4.x? I didn't see it called anymore unfortunately

tinovyatkin:
Templates can be overriden by providing `callbackOnCreateTemplates`, that saves merged object in `this.config.templates`. However, while refactoring for 4.x, `WrappedSelect` was hard bounded to original, non user provided callbacks. This PR fixes that also fixes #461.


jshjohnson:
Thanks 🎉

PR #669, opened 2019-10-21
Improve GitHub actions and fix `package.json` SCSS script

tinovyatkin:
1. Improves GitHub actions to fore run on dependencies and browserlistrc changes
2. Add missed commit information to BundeSize run and token for BundleSize run on master. We actually don't need to rebuild minified assets as currently they are part of repository. I would suggest to remove them from repository and just build current version on publish.
4. Updated `node-sass` to latest version to trigger this actions run on this PR and also allow installing on Node 12.13 (new LTS starting today :tada:)
5. Upgraded `csso` dependency to correct `csso-cli` that cuts sourcemaps (to let `budnlesizes` pass)
6. Fixes `run-p` use instead of `run-s` at `css-build`


jshjohnson:
I've added the `CYPRESS_RECORD_KEY` secret 👍

.github/workflows/bundlesize.yml
```diff
@@ -5,6 +5,8 @@ on:
     paths:
       - 'src/scripts/**'
       - 'src/styles/**'
+      - 'package-lock.json'
```
jshjohnson: Nice 👍

jshjohnson:
>  I would suggest to remove them from repository and just build current version on publish.

jshjohnson:
>  I would suggest to remove them from repository and just build current version on publish.

jshjohnson:
> I would suggest to remove them from repository and just build current version on publish.

jshjohnson:
> I would suggest to remove them from repository and just build current version on publish.

I'd ideally like to do this but they're required for the demo page (deployed to gh-pages). If there is a clever way of deploying to building assets and deploying them to gh-pages, that'd be the best case scenario.

tinovyatkin:
> I've added the `CYPRESS_RECORD_KEY` secret 👍

@jshjohnson  Are you sure there is no error (in name, empty value, etc)? It doesn't picks the value...

tinovyatkin:
> > I would suggest to remove them from repository and just build current version on publish.
> 
> I'd ideally like to do this but they're required for the demo page (deployed to gh-pages). If there is a clever way of building assets and deploying them to gh-pages, that'd be the best case scenario.

@jshjohnson well, GitHub Actions is a clever way :-)
You can always do something like after build (I'm just typing here, so, need to try :-))
```yml

run: |
           git stash
           git checkout gh-pages
           git stash pop
           git add public/assets/scripts/choices.min.js public/assets/styles/base.min.css public/assets/styles/choices.min.css
           git commit -m "Updating productions assets"
           git push https://${GITHUB_ACTOR}:${{ secrets.GITHUB_TOKEN }@github.com/${GITHUB_REPOSITORY}.git HEAD:gh-pages
```

tinovyatkin:
@jshjohnson remove Cypress recording from this PR, so, you can merge it and check the CYPRESS_RECORD_KEY later on, as other changes here are more important

jshjohnson:
hey @tinovyatkin - I've just merged your other PR and got this error when publishing to NPM:

```npm ERR! 403 Forbidden - PUT https://registry.npmjs.org/choices.js - You cannot publish over the previously published versions: 7.1.1.```

jshjohnson:
hey @tinovyatkin - I've just merged your other PR and got this error when publishing to NPM:

```npm ERR! 403 Forbidden - PUT https://registry.npmjs.org/choices.js - You cannot publish over the previously published versions: 7.1.1.``` 

I was under the impression that it would only attempt to run the "Publish Package" workflow if a tag had been pushed with the PR? Or am I just misunderstanding

tinovyatkin:
> hey @tinovyatkin - I've just merged your other PR and got this error when publishing to NPM:
> 
> `npm ERR! 403 Forbidden - PUT https://registry.npmjs.org/choices.js - You cannot publish over the previously published versions: 7.1.1.`
> 
> I was under the impression that it would only attempt to run the "Publish Package" workflow if a tag had been pushed with the PR? Or am I just misunderstanding

It does run on both, pushes to master or pushes of tags starting with `v`. If version in `package.json` wasn't changed then failed attempt to publish can be just ignored. The whole on `master` however, is always makes sense, as it uploads code coverage and bundlesizes base sizes to compare. We can separate in two actions or add more conditions to the run, but it doesn't hurts to leave as is.

tinovyatkin:
@jshjohnson 🏁

PR #674, opened 2019-10-22
Increase Unit test coverage by testing SELECT case

tinovyatkin:
1. This PR increases unit test coverage for `choices.js` from 38.91% to 54.69% and total coverage for statements from 55.27% to 66.62% 🎉   by testing small use case from my own practice.

2. Add global `Option` object to JSDom mocking.

PS: Something happened with this repository secrets and Codecov can't find it's token too now :-(

### DIFF
--- a/src/scripts/choices.test.js
+++ b/src/scripts/choices.test.js
@@ -51,6 +51,48 @@ describe('choices', () => {
         });
       });
 
+      describe('init with SELECT element', () => {
+        it('works with empty select and choices array', () => {
+          const select = document.createElement('select');
+          document.body.appendChild(select);
+          const calledValues = [];
+          const choices = new Choices(select, {
+            choices: [
+              {
+                value: 'Option 1 value',
+                label: 'Option 1',
+                selected: false,
+              },
+              {
+                value: 'Option 2 value',
+                label: 'Option 2',
+                selected: true,
+              },
+            ],
+            callbackOnCreateTemplates: () => ({
+              option({ label, value, active }) {
+                calledValues.push(value);
+                return new Option(label, value, false, active);
+              },
+            }),
+          });
+          expect(choices.getValue(true)).to.equal('Option 2 value');
+          expect(select.value).to.equal('Option 2 value');
+          expect(select.childElementCount).to.equal(1);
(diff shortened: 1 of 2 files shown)


src/scripts/choices.test.js
```diff
@@ -51,6 +51,48 @@ describe('choices', () => {
...
+
+          choices.setChoiceByValue('Option 1 value');
+          expect(calledValues).to.deep.equal([
+            'Option 2 value',
+            'Option 2 value',
+            'Option 1 value',
+            'Option 1 value',
+          ]);
+        });
+      });
```
jshjohnson: Thanks but i'm reluctant to merge this as tests like this aren't really testing anything specific. They're just covering a lot of ground which increases coverage. If this test fails, it wouldn't be immediately clear what was actually broken. 

To be honest the main choices.js file needs splitting out/rewriting to make testing easier. Once that's done, increasing coverage will be easier

tinovyatkin: I agree with you regarding needs for `choices.test.js` refactoring (I'm actually was thinking about submitting a PR migrating all unit tests to Jest, instead of current `mocha` + `sinon` + `chai` + `jsdom` + `nyc`, but not sure you will accept that before investing my time). But in the meantime I think the current unit coverage for `choices.js` is extremely low, and while integration tests with Cypress are more extensive, it helps a lot to have faster feedback from unit tests.

However, I can't accept that this test is redundant and testing already covered functionality (in this case it would not increase total coverage by 16%).

1. That's the only test I've found where you are creating `Choices` instance for HTMLSelectElement

2. That's the only test I've found where you are creating Choices by passing array of items as `choices` parameter to constructor.

3. That's the only test I've found that test actual use of templates creation callback

4. That's the only test testing the fact that original SELECT element should be updated on `setChoiceByValue` (you are not testing this functionality at related test suite).

I can split all above into 4 unit tests, but it will result in repeating the same code 4 time and instead of 1 additional test of 15 lines we will have 4 more in the already clogged `choices.test.js`

If this test fails, first of all, it will point to a new bug, and in additional you will have a stack trace to _trace_ ;-)

BONUS: I've added `Option` to globals :-)

PR #682, opened 2019-10-23
Fixing typings (again 🤷🏼‍♂️) + test for them

tinovyatkin:
Fixing two small misspellings + added GitHub action to prevent that in the future.

There is a bigger problem however - Choices constructor is little hacky and can return one instance of Choices (like any normal class), or array of Choices (if created with string parameter) or even `undefined` (if created with unsupported parameter). As far I know, it's impossible to describe that in TypeScript, but I've added a small hack to at least allow it to use today.

### DIFF
--- a/.github/workflows/types.yml
+++ b/.github/workflows/types.yml
@@ -1,0 +1,27 @@
+name: TypeScript Check
+
+on:
+  pull_request:
+    paths:
+      - 'types/index.d.ts'
+
+jobs:
+  tsc:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


jshjohnson:
> There is a bigger problem however - Choices constructor is little hacky and can return one instance of Choices (like any normal class), or array of Choices (if created with string parameter) or even undefined (if created with unsupported parameter). As far I know, it's impossible to describe that in TypeScript, but I've added a small hack to at least allow it to use today. 

Yeah agreed, not my best decision! Will have to revisit that

PR #683, opened 2019-10-23
breaking test - constructor returning multiple instances

tinovyatkin:
After realising  hacky nature of Choices constructor (#682)  I've decided to try it a little this way and created this test, that currently is breaking. @jshjohnson  Is there something wrong in the test or in Choices?

### DIFF
--- a/src/scripts/choices.test.js
+++ b/src/scripts/choices.test.js
@@ -15,6 +15,22 @@ describe('choices', () => {
     });
   };
 
+  describe('constructor can return array of instances', () => {
+    it('should be the same as number of matched elements', () => {
+      document.body.innerHTML = `
+      <input data-choice type="text" id="inp1" />
+      <input data-choice type="text" id="inp2" />
+      <input data-choice type="text" id="inp3" />
+      `;
+
+      const inputs = document.querySelectorAll('[data-choice]');
+      expect(inputs.length).to.equal(3);
+
+      const choices = new Choices();
+      expect(choices.length).to.equal(inputs.length);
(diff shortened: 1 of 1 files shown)


src/scripts/choices.test.js
```diff
@@ -15,6 +15,22 @@ describe('choices', () => {
...
+      document.body.innerHTML = `
+      <input data-choice type="text" id="inp1" />
+      <input data-choice type="text" id="inp2" />
+      <input data-choice type="text" id="inp3" />
+      `;
+
+      const inputs = document.querySelectorAll('[data-choice]');
+      expect(inputs.length).to.equal(3);
+
+      const choices = new Choices();
```
jshjohnson:
```suggestion
      const choices = new Choices(inputs);
``` 
:)

tinovyatkin: The whole point here is to construct from a string. If a pass nothing to Choices constructor then defaults is a string ‘[data-choice]’

tinovyatkin: By the way, `new Choices(inputs)` will not work - Choices is not supporting directly `NodesList` as constructor parameter.

tinovyatkin: The thing is that if the constructor never worked in the expected way with selector returning several elements, then you can adjust/fix its behaviour in a minor version as it cannot be considered as breaking change.

jshjohnson:
@tinovyatkin see https://github.com/jshjohnson/Choices/pull/693 👍

PR #685, opened 2019-10-23
set window target for minified build

tinovyatkin:
This PR doing 3 important things:

1. Taking in account that it's a browser-only library setting `globalObject: 'window'` for webpack

2. Repointing `main` from `package.json` to non-minified UMD version and ships minified version as bounded to `window` function, not as UMD module.
Motivation here: if some developer are using this library via `import Choices from 'choices.js'` then that developer expect are non-minified version of the library to become part of his bundle (as they may  debug something or apply different minifier later, for example Closure Compiler and existing minifying will hurt). This expectation particularly set by common sense and by [the most popular Javascript library of all times](https://github.com/jquery/jquery/blob/bcbcdd2b2c1bb7075f4f73dc89ca7d65db2a09ed/package.json#L6).
As result, coupled with upgraded `fuse.js` it decreases minified bundle size.

3. Committed affected minified builded asset (we should get rid of them from repository, seriously :-) )


gregjacobs:
@jshjohnson Just hitting an issue with regards to this: React apps that include Choices which are server-side rendered throw an error because `window` is not defined on the server.

What was the issue with leaving the default global var for UMD?

PR #687, opened 2019-10-24
Remove classnames and use `createElement`

tinovyatkin:
I know there is a lot of religious wars on `innerHTML` vs `createElement` approach, so, not expecting this PR will pass easily.

But I see several benefits in the creating direct DOM elements in templates instead of using `strToEl` helper:

1. While for complex DOM structure innerHTML may be faster, most of templates creating in this library is just one element, so, it's overkill to first create complex string template, attach it to a temporary DIV element, get from it as first child, remove child. I'm pretty sure direct approach is faster for this use case.

2. It shorter (almost 50 lines shorter) and VSCode Interllisense works to help, contrary to pure text templates, that are human errors prone.

3. One dependency less, `classnames` gone (it's small, but).

4. Resulting JS bundle is smaller now - ~18 kb

Also, there are two problems I've found doing this PR:

- `item` has a hard coded English string and the only way to localize it - is to overwrite whole template.
- the same `item` template omits customProperties when `removeItemButton` is false, but renders them otherwise

This code can be further improved by creating a helper function, but it will decrease readability, so, I prefer keep as it now.


src/scripts/templates.js
```diff
@@ -1,238 +1,192 @@
...
-  item(globalClasses, data, removeItemButton) {
-    const ariaSelected = data.active ? 'aria-selected="true"' : '';
-    const ariaDisabled = data.disabled ? 'aria-disabled="true"' : '';
 
-    let localClasses = classNames(globalClasses.item, {
-      [globalClasses.highlightedState]: data.highlighted,
-      [globalClasses.itemSelectable]: !data.highlighted,
-      [globalClasses.placeholder]: data.placeholder,
+  item(
+    { item, button, highlightedState, itemSelectable, placeholder },
```
jshjohnson: Subjective thing but I think this almost makes the code less readable as it's not immediately obvious their class names

src/scripts/templates.js
```diff
@@ -1,238 +1,192 @@
...
+      id,
+      value,
+      customProperties,
+    });
+    if (active) div.setAttribute('aria-selected', 'true');
+    if (disabled) div.setAttribute('aria-disabled', 'true');
+
+    if (isPlaceholder) div.classList.add(placeholder);
+    if (highlighted) div.classList.add(highlightedState);
+    else div.classList.add(itemSelectable);
```
jshjohnson: Can we do normal if/else blocks please :) surprised eslint didn't moan at this

src/scripts/templates.js
```diff
@@ -1,238 +1,192 @@
...
+  notice({ item, itemChoice, noResults, noChoices }, innerHTML, type = '') {
+    return Object.assign(document.createElement('div'), {
+      innerHTML,
+      className: `${item} ${itemChoice} ${
+        // eslint-disable-next-line no-nested-ternary
+        type === 'no-results'
+          ? noResults
+          : type === 'no-choices'
+          ? noChoices
+          : ''
```
jshjohnson: Think this would be cleaner if it was stored in a variable and referenced 👍

src/scripts/templates.js
```diff
@@ -1,238 +1,192 @@
...
-      <option value="${data.value}" ${data.active ? 'selected' : ''} ${
-      data.disabled ? 'disabled' : ''
-    } ${
-      data.customProperties
-        ? `data-custom-properties=${data.customProperties}`
-        : ''
-    }>${data.label}</option>
-    `);
+  option({ label, value, customProperties, active, disabled }) {
+    const opt = new Option(label, value, false, active);
```
jshjohnson: Is there a reason why we don't use `createElement('option')` here? Or alternatively `new HTMLDivElement(...)` above? 🤔

src/scripts/templates.test.js
```diff
@@ -40,11 +56,7 @@ describe('templates', () => {
             searchEnabled,
             passedElementType,
           );
-
-          expect(getType(actualOutput)).to.equal('HTMLDivElement');
-          expect(stripElement(actualOutput)).to.equal(
-            stripElement(expectedOutput),
-          );
+          expectEqualElements(actualOutput, expectedOutput);
```
jshjohnson:
```suggestion
          expectEqualElements(expectedOutput, actualOutput);
```

jshjohnson: If you have them this way round, all the attributes of the expected output will be checked as you're looping over `element1.attributes` above

src/scripts/templates.test.js
```diff
@@ -491,29 +458,31 @@ describe('templates', () => {
     };
 
     it('returns expected html', () => {
+      /*
+        Following attributes are not supported by JSDOM, so, can't compare
+          autocapitalize="off"
+          spellcheck="false"
+      */
```
jshjohnson: 👍

src/scripts/templates.js
tinovyatkin: Shorter code using standard constructor that also sets attributes (selected and value) immediately

src/scripts/templates.test.js
tinovyatkin: I’m comparing first that two elements have the same number of attributes - after that it’s not important which one to loop

src/scripts/templates.js
tinovyatkin: Do you mean ‘{}’? Prettier config removing that rule that why no warning here. Will do.

src/scripts/templates.js
tinovyatkin: regarding `new HTMLDivElement` - there is no such constructor.

tinovyatkin:
Will add JSDoc typing to help with destruction parameters and rebase over update master. Please don't merge yet.

src/scripts/templates.js
tinovyatkin: Yes, in that particular case may be, but I'd keep it like this just for uniformity with the rest of callbacks here.

tinovyatkin:
@jshjohnson Added strict typing via JSDocs, so, VSCode now know types of every parameter and will catch any misspeling in destructions or return values.

Had to refactor/fix typings again, removed all (outdated anyway) private APIs from there, fixed the rest of types, and was not able to resist to add myself as an author of the beauty to be blamed by future generations ;-)

Prefer not to fix much of stylistic things here, as it counter logic in projects that are using Prettier and I'm preparing a PR to unleash its potential better here.

That's pretty much 🏁

tinovyatkin:
@jshjohnson are this going to be merged? I would like to fix #689 in base of this one. Please let me know if you something left to refactor here.

jshjohnson:
> @jshjohnson are this going to be merged? I would like to fix #689 in base of this one. Please let me know if you see something left to refactor here.

Done!

PR #688, opened 2019-10-24
Remove built assets from git

jshjohnson:


jshjohnson:
@tinovyatkin I can't find your comment regarding having the `npm deploy` script run as a github action. Is this technically possible?

It would need to:
- Only deploy after a successful npm publish
- Build assets
- Push the contents of the `public` folder to the `gh-pages` branch (including all built assets)

The inclusion of the built assets is the tricky bit 🤔

jshjohnson:
This is what I based the current setup on:  https://gist.github.com/cobyism/4730490

tinovyatkin:
> @tinovyatkin I can't find your comment regarding having the `npm deploy` script run as a github action. Is this technically possible?

The original comment is here: https://github.com/jshjohnson/Choices/pull/669#issuecomment-545031231

Yes, it's possible to run `npm deploy` from GitHub Action, but if you are doing `git push` from that action then you need to specify full repository URL with authentication parameters, as in my comment.
GitHub is [making things easier](https://github.com/actions/checkout/pull/56) regarding `git push` from Actions, but it's not released yet.

PR #690, opened 2019-10-24
Add deploy to gh pages action

jshjohnson:


tinovyatkin:
@jshjohnson 
I would suggest for test to separate it in an indepent action that runs on master on a specific file path (.gitignore?) (`- path`). Then push that file to master to check publish running well.

PR #693, opened 2019-10-24
Fix constructor

jshjohnson:
## Description
As highlighted by @tinovyatkin [here](https://github.com/jshjohnson/Choices/pull/683), the ability to pass multiple elements to the Choices constructor and get multiple instances back was buggy at best. This PR removes that functionality so the Choices constructor will only ever return one instance and that instance will only be associated with one element. To achieve the same behaviour, you would need to do the following:

```js
const elements = Array.from(document.querySelectorAll('.your-class'));

const instances = elements.map((element) => {
  return new Choices(element);
});
``` 

or even simpler:

```js
const elements = Array.from(
  document.querySelectorAll(".your-class"),
  element => new Choices(element)
);
```

I've also updated the constructor to throw an error if no valid element is passed _or_ no valid element is found in the DOM. Previously this would have failed silently. 

## How Has This Been Tested?
Constructor tests have been added with this change.

## Types of changes

## Checklist:


README.md
```diff
@@ -53,14 +53,11 @@ Or include Choices directly:
...
-will be returned. If you target one element, that instance will be returned.
+**Note:** If you pass a selector which targets multiple elements, the first matching element will be used. Versions prior to 8.x.x would return multiple Choices instances.
 
 ```js
-  // Pass multiple elements:
-  const [firstInstance, secondInstance] = new Choices(elements);
-
-  // Pass single element:
+  // Pass single element
+  const element = document.querySelector('.js-choice')
```
jshjohnson:
```suggestion
  const element = document.querySelector('.js-choice');
```

tinovyatkin:
@jshjohnson `Array.from` accepts second parameter, so, code may be easier:

```js
const elements = Array.from(
  document.querySelectorAll(".your-class"),
  element => new Choices(element)
);
```

tinovyatkin:
@jshjohnson consider also #691 for 8x 🙏

jshjohnson:
> Also need to update typings by removing brackets hack.
> And probably document behaviour if Choices is constructed with selector that may return several elements (we use `document.querySelector` to pick the first one)

See here: https://github.com/jshjohnson/Choices/pull/693/commits/4435efc5afa799ea033299c7c0bb9fd28f3aef40 - let me know if this is wrong

src/scripts/choices.js
```diff
@@ -74,19 +65,18 @@ class Choices {
...
 
-    if (!passedElement) {
-      if (!this.config.silent) {
-        console.error(
-          'Could not find passed element or passed element was of an invalid type',
-        );
-      }
+    if (isType('String', element)) {
+      passedElement = document.querySelector(element);
+    } else if (isElement(element)) {
```
tinovyatkin: If you will use here directly `if (element instanceof HTMLInputElement || element instanceof HTMLSelectElement)` instead of `isElement` helper it will you give well better typed autocomplete in VSCode

src/scripts/choices.js
```diff
@@ -74,19 +65,18 @@ class Choices {
...
+      passedElement = element;
+    }
 
-      return;
+    if (!passedElement) {
+      throw Error(
```
tinovyatkin: throw `TypeError` here as more appropriate

src/scripts/choices.js
```diff
@@ -1787,7 +1779,7 @@ class Choices {
       userTemplates = callbackOnCreateTemplates.call(this, strToEl);
     }
 
-    this.config.templates = merge(TEMPLATES, userTemplates);
+    this.templates = merge(TEMPLATES, userTemplates);
```
tinovyatkin: I believe `templates` should be private member (as consumer should not have direct access to it)? So, something like `_templates` by your notation...

PR #694, opened 2019-10-25
don't calculate pixel width of input

tinovyatkin:
I'm probably missing something, as there is no visual test for this functionality, but.
You currently have a great deal of code to calculate exact pixel width of helper INPUT element and it's doing good job. However, I think that _almost the same functionality_ with little bit better compatibility with accessibility displays and other rare cases, as well as with **much less code** can be obtained by using [well supported](https://caniuse.com/#feat=ch-unit) `ch` unit.
`ch` is a width of `0` in current font and it's not _exact_ measurements of value string width, but you are using `pre` and _magic number_ of 75% in this function anyway, so this variant can be tuned in similar way (something like `length * 0.75` [should be pixel perfect](https://meyerweb.com/eric/thoughts/2018/06/28/what-is-the-css-ch-unit/#targetText=I%20keep%20seeing%20authors%20and,a%20certain%20number%20of%20characters!%E2%80%9D))  if you want. For my taste, I will leave it like this.

Also found that current tests are using non-existent `_placeholderValue` property, but still passing well 🧐

Unfortunately, this functionality can't be tested with JSDOM due to lack of ch support there (I've filled [appropriate Pull Request](https://github.com/jsdom/cssstyle/pull/107)), so, I've tested by temporary using `em` units, and also manually in browser.

Now, it's up to you, as it's not pixel perfect and have some edge case where it leaves a little of space at the end on some strings. But I ❤️ _this_ solution.

### DIFF
--- a/src/scripts/components/input.js
+++ b/src/scripts/components/input.js
@@ -1,4 +1,4 @@
-import { calcWidthOfInput, sanitise } from '../lib/utils';
+import { sanitise } from '../lib/utils';
 
 export default class Input {
   constructor({ element, type, classNames, placeholderValue }) {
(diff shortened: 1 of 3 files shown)


tinovyatkin:
My Pull Request to JSDOM/cssstyle was merged tonight, so, soon it will be also testeable feature (with next JSDOM release hopefully).

jshjohnson:
This is awesome 👌 thanks

PR #696, opened 2019-10-25
[MAJOR] don't reimplement native `closest`

tinovyatkin:
Not sure if new polyfill is a major semver change, but just in case.
As we already in game with polyfills it makes no sense to reimplement native `Element.prototype.closest` here.


tinovyatkin:
@jshjohnson need to add `Element.prototype.closest` also to settings in #709, if that one will be merged first

PR #697, opened 2019-10-25
remove redundant CSS

tinovyatkin:
Removing leftovers of #691 - that elements are already hidden by `hidden` attribute.

Not sure how the fix in JS are related here (but it was a bug anyway), but without these changes tests are breaking.


jshjohnson:
Thanks again for your contributions!! 😄 💯

PR #698, opened 2019-10-25
use default parameters

tinovyatkin:
Small refactoring to use default parameter instead of `||` in function body and one removal of excessive array spread.

### DIFF
--- a/src/scripts/choices.js
+++ b/src/scripts/choices.js
@@ -390,12 +390,15 @@ class Choices {
     return this._isSelectOneElement ? values[0] : values;
   }
 
-  setValue(args) {
+  /**
+   * @param {string[] | import('../../types/index').Choices.Item[]} items
+   */
+  setValue(items) {
     if (!this.initialised) {
       return this;
     }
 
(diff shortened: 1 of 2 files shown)


src/scripts/choices.js
```diff
@@ -390,12 +390,15 @@ class Choices {
     return this._isSelectOneElement ? values[0] : values;
   }
 
-  setValue(args) {
+  /**
+   * @param {string[] | import('../../types/index').Choices.Item[]} items
+   */
+  setValue(items) {
```
jshjohnson: Would you mind updating the README too please :)

tinovyatkin: Sorry, didn’t get this one - README already says that setValue parameter is always an array, here I’m removing unneeded spread+reconstruction of parameter and added parameters description to prevent that in future

jshjohnson: It just says "args" in the README though - it should be "items" - https://github.com/jshjohnson/Choices#setvalueargs

PR #699, opened 2019-10-25
rename addItemFilterFn

tinovyatkin:
## Types of changes

## Checklist:

### DIFF
--- a/src/scripts/choices.js
+++ b/src/scripts/choices.js
@@ -62,6 +62,18 @@ class Choices {
       { arrayMerge: (destinationArray, sourceArray) => [...sourceArray] },
     );
 
+    // Convert addItemFilter to function
+    if (
+      userConfig.addItemFilter &&
+      typeof userConfig.addItemFilter !== 'function'
+    ) {
+      const re =
+        userConfig.addItemFilter instanceof RegExp
+          ? userConfig.addItemFilter
+          : new RegExp(userConfig.addItemFilter);
+      this.config.addItemFilter = re.test.bind(re);
+    }
+
     const invalidConfigOptions = diff(this.config, DEFAULT_CONFIG);
     if (invalidConfigOptions.length) {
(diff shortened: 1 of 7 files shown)


src/scripts/choices.js
```diff
@@ -62,6 +62,18 @@ class Choices {
       { arrayMerge: (destinationArray, sourceArray) => [...sourceArray] },
     );
 
+    // Convert addItemFilter to function
+    if (
+      userConfig.addItemFilter &&
+      typeof userConfig.addItemFilter !== 'function'
```
jshjohnson: Using `isType('Function', this.config.addItemFilter)` would be consistent with the rest of the codebase. I've avoided using typeof due to this: https://bonsaiden.github.io/JavaScript-Garden/#types.typeof

tinovyatkin: I will submit a PR later on to remove isType, etc. If you are going to TypeScript you probably will do it yourself too.

`typeof` together with `instanceof` is a major tool TypeScript/VSCode type Intellisence relays on.  It's the [type guards](https://ultimatecourses.com/blog/understanding-typescript-instanceof-type-guard). While they have an edge cases (like `typeof new String('who is doing that?!') !== typeof String('🤷🏼‍♂️')`, for simply checking that something is a function `typeof fn === 'function'` is flawless as well as `r instanceof RegExp` and the last one helps a lot to prevent further developer error with automatic type checking.

tinovyatkin: @jshjohnson by the way, here are some breaking test for your current implementation of `isType`:

```js
expect(isType("Function", async () => {})).to.equal(true);
expect(
  isType("Function", function* gen() {
    yield false;
  })
).to.equal(true);
```

jshjohnson: Fair enough - in that case it's probably worth just ditching `isType`

PR #700, opened 2019-10-25
[Tooling] Transpile tests targeting Node, not browser

tinovyatkin:
Just found we are transpiling test files using browser target, while we actually may do faster Node target. Fixes that.
Found a bug in test as result of this and also fixed that.


tinovyatkin:
@jshjohnson 🙏 need that to have `async` functions in unit tests...

PR #703, opened 2019-10-25
remove delegateEvents, use passive event handlers, remove leftovers

tinovyatkin:
This PR removes `window.delegateEvents` use in favour of regular and passive event handlers.
Also removing orphaned form reset handler from Input class.

Motivation here:
While I do believe I understand why `delegateEvents` was added in the first place (_to have global capture events shared between all page instances of Choices, correct?_ 🙋🏻‍♂️) I believe this solution is counter productive and should be refactored:

1. It's pretty easy to get a name collision, as `window.delegateEvents` sound very generic and nowhere documented. You may be safe there with a `Symbol.for`, but it's not supported in IE11. If the name collision even happens and some other library overwrites this name you will have to deal with very difficult debuging

2. You still have event handler code in each instance, so, no memory optimisation here.

3. You basically are repeating what standard `addEventListener` doing. Save event handling functions -> dispatch them. No magic, no savings here.

4. You are listening for some very common events (like `touchmove` and `mouseover`) in `capture` mode while it's just observing and passive mode will be fine and more performant.

Altogether, while this microoptimization may have some theoretical sense, it's definitely makes code more error prone and the Choices library way less _good citizen_ from page **overall performance** point of view.


src/scripts/choices.js
```diff
@@ -1085,48 +1084,91 @@ class Choices {
...
-    window.delegateEvent.add('mousedown', this._onMouseDown);
-    window.delegateEvent.add('mouseover', this._onMouseOver);
+    const { documentElement } = document;
+
+    // capture events - can cancel event processing or propagation
+    documentElement.addEventListener('keydown', this._onKeyDown, true);
+    documentElement.addEventListener('touchend', this._onTouchEnd, true);
+    documentElement.addEventListener('mousedown', this._onMouseDown, true);
+
+    // passive events - doesn't call preventDefault of stropPropagation
```
```suggestion
    // passive events - doesn't call preventDefault or stropPropagation
```

PR #705, opened 2019-10-26
[Tooling] VSCode Debug configurations and project settings

tinovyatkin:
This PR aims to make a contributor life easier by providing ready environment in VScode:

1. **Debug configurations** 
image
Provided ready to use debugging settings:
 - **Launch Chrome** - fires development server in background, and launches Chrome with configured source map support, so, things like breaking points works out of the box. Adjusted webpack configuration here to build source maps in development
image

- **Mocha Current File** - allow to start debugging session from opened Mocha *.test.js file with all the breaking points, etc.
Screenshot 2019-10-26 at 17 18 07

- **Mocha All files** - the same, but run all suites, so, yo can start, for example, from a source file by setting a breakpoint.

- **Cypress Current File** _(that was the trickiest one 😫 )_ - from an opened `*.spec.js` fires development server (if it's not online already) and launches Cypress in headed and non-exit mode and allows to [unleash full debugging](https://docs.cypress.io/guides/guides/debugging.html)
image

2. **Tasks** - configured tasks and problem marchers for `package.json` tasks

3. **Project settings** - added further settings to prevent Prettier and built-in formatters conflicts, improve Intellisense on some files, and enforce file formats for Window-based contributors.

4. **Adjusted `jsconfig.json`**. Most important change here is `checkJs: true` - you will start to see some annoying false positive errors from VSCode (until we will have JSDoc typing everywhere), but it helps _a lot_ with catching some hard to find errors, like this one (fixed in #694) :
Screenshot 2019-10-26 at 17 02 37

Let's do Visual Studio Code great again, try CMD + Shift + B 😄

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,4 +1,0 @@
-module.exports = {
-  singleQuote: true,
-  trailingComma: "all"
-}
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,0 +1,13 @@
+{
+  "singleQuote": true,
+  "trailingComma": "all",
+  "overrides": [
+    {
+      "files": ["*.svg"],
+      "options": {
+        "parser": "html",
+        "htmlWhitespaceSensitivity": "ignore"
+      }
+    }
+  ]
+}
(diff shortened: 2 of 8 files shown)


jshjohnson:
Amazing 🎉 thanks

PR #707, opened 2019-10-26
fix preventPaste

tinovyatkin:
There is a reference to `Input.preventPaste` property in the Input element at [`onPaste` handler](https://github.com/jshjohnson/Choices/blob/15d54c7d342f5e44ad2ccf8a3ec5b81e56752a4c/src/scripts/components/input.js#L134). But it's not set anywhere. Is it missing functionality or some leftovers?

`classNames` also have no use in this class, do you want me to remove them? 

`placeholderValue` also is not on the call in [the only place where you creating an instance of Input](https://github.com/jshjohnson/Choices/blob/15d54c7d342f5e44ad2ccf8a3ec5b81e56752a4c/src/scripts/choices.js#L1824) element. It should either be renamed to `placeholder` and added to constructor call or (coupling with #694) removed at all.

### DIFF
--- a/src/scripts/choices.js
+++ b/src/scripts/choices.js
@@ -1825,6 +1825,7 @@ class Choices {
       element: this._getTemplate('input', this._placeholderValue),
       classNames: this.config.classNames,
       type: this.passedElement.element.type,
+      preventPaste: !this.config.paste,
     });
 
     this.choiceList = new List({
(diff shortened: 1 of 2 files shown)


tinovyatkin:
@jshjohnson is that related to https://github.com/jshjohnson/Choices#paste ?

jshjohnson:
This looks like a regression bug - it should prevent a user pasting into the input. It just needs to be passed here: https://github.com/jshjohnson/Choices/blob/master/src/scripts/choices.js#L1824-L1828

Issue #637, opened 2019-09-17
selectedState does not exist


St-Permiakov:
Related to #367 
As stated there, selectedState css-class does not appear on selected option.
So either documentation is incorrect or it's a bug.

mroseboom:
Hey there, any update on this?

tinovyatkin:
@jshjohnson Do you want that to be implemented or removed from documentation?

jshjohnson:
I think it can be removed - I don't see any distinction between this and the `highlighted` state

St-Permiakov:
It most assuredly cannot.
Highlighted state appears on hover and selected should appear on currently selected items. Those are completely different use cases.

jshjohnson:
@St-Permiakov you could use the 'highlighted' state for that situation

mroseboom:
@jshjohnson At this moment it isn't implemented or i'm missing someting?

tinovyatkin:
@mroseboom it's not implemented in code base in current version.

St-Permiakov:
Does it appear on selected item without user hovering it?
How can I distinguish selected item when I'm hovering it?
And what if I'm supposed to make selected item visually disabled (that was my use case actually)?

tinovyatkin:
@jshjohnson the implementation will be trivial and I can easily add that, however, @St-Permiakov use case probably will anyway require overriding template rendering. 

Is that feature were never implemented or were removed in past version without updating documentation?

St-Permiakov:
Well, the only reason why I created this issue was impossibility to distinguish hovered and selected states for styling. That's what that selectedState was supposedly for. If there is any other way...

tinovyatkin:
@St-Permiakov with Choices 8 released today:

```js
const example = new Choices(element, {
  callbackOnCreateTemplates: () => ({
    item: (classes, item) =>
      const i = Choices.defaults.templates.input.call(this, classes, item);
     if(item.selected) i.classList.add('.is-selected');
return i;
  }),
});
```

stasguma:
@tinovyatkin For me works the second example with a `choice` function. Thanks a lot.
I'll correct you a little.
In this part `classList.add(".is-selected")` we don't need a dot before `is-selected`.

It would be better if it worked without a workaround

tinovyatkin:
@stasguma glad it works for you, fixed an example, thank you!

jshjohnson:
[...]

PR #708, opened 2019-10-26
[MAJOR] expose `defaults` and original Template rendering functions

tinovyatkin:
In most real life use cases for `callbackOnCreateTemplates` (like #637), consumer actually want to extend original render a little, without fully overwriting it. However, there is no easy way to do that.
This PR exposes original rendering functions as static readonly property `Choices.defaults.templates`, so, following code will become possible:

```js
const example = new Choices(element, {
  callbackOnCreateTemplates: () => ({
    input: (...args) =>
      Object.assign(Choices.defaults.templates.input.call(this, ...args), {
        type: "email"
      })
  })
});
```

MAJOR because it moving undocumented `userDefaults` into `Choices.defaults`


src/scripts/choices.js
```diff
@@ -177,6 +177,14 @@ class Choices {
     this.init();
   }
 
+  /* ========================================
+  =            Static properties            =
+  ======================================== */
+
+  static get Templates() {
+    return TEMPLATES;
+  }
```
jshjohnson: 👍 Can we define `userDefaults` here too please? See https://github.com/jshjohnson/Choices/blob/master/src/scripts/choices.js#L2095 

Is there a reason for this property not being upper cased?

tinovyatkin: `userDefaults` are mutable, so, we will need to introduce extra variable to keep it value. Are you ok with that? It seems more logical to have it as static property, but.

I don't know your conventions for this lib regarding static properties, what would you prefer - Choices.TEMPLATES or Choices.TEMPLATES? I think it should be more or less in line with userDefaults?

jshjohnson: I definitely like the form of `Choices.defaults.xyz` 👍

jshjohnson: This will be going into the next major release anyway so we can have breaking changes here

src/scripts/choices.js
```diff
@@ -39,11 +39,28 @@ import {
...
+  static get defaults() {
+    return Object.preventExtensions({
+      get options() {
+        return USER_DEFAULTS;
+      },
+      get templates() {
+        return TEMPLATES;
+      },
+    });
+  }
```
jshjohnson: 👌

PR #709, opened 2019-10-27
[Tooling] Expand browsers target while keeping Polyfills under control

tinovyatkin:
This PR is aims to get more out of `.browserslistrc` file while ensuring we have automated checks in place to prevent undocumented/non-polyfilled feature use.

1.  Expands (and simplifies) browsers target to just `> 1%`. This is wider than your current `> 5% and IE11`, and includes, for example, Firefox 69 and Edge 18. From JS point of view it doesn't change anything (as IE11 was lower target), but from CSS point (used via Autprefixer) it adds a couple of prefixes more.

2. Adds and configures with expected polyfills [`eslint-plugin-compat`](https://github.com/amilajack/eslint-plugin-compat). We use it as regular plugin and also at GitHub action to lint resulting bundle and ensure it stays within ES5 and documented polyfills.

3. Added automatic check of sync between polyfills documented in README and declared for `eslint-plugin-compact`. Also updated documentation links to version 3 of polyfill.io


.github/actions-scripts/polyfills-sync.js
```diff
@@ -0,0 +1,14 @@
...
+const readme = readFileSync(path.resolve(__dirname, '../../README.md'), 'utf8');
+
+const polyfillsFromDocs = /^```polyfills\s*\n([^`]+)\n^```/m
+  .exec(readme)[1]
+  .split('\n')
+  .map(v => v.trim())
+  .sort();
+// @ts-ignore
+const polyfillsFromSettings = require('../../.eslintrc.json').settings.polyfills.sort();
+assert.deepStrictEqual(polyfillsFromDocs, polyfillsFromSettings);
```
jshjohnson: This is cool! 😃

public/assets/scripts/.eslintrc.js
```diff
@@ -0,0 +1,24 @@
...
+  ],
+  parserOptions: {
+    // ensure that it's compatible with ES5 browsers, so, no `const`, etc
+    ecmaVersion: 5
+  },
+  env: {
+    browser: true
+  },
+  settings
+})
```
jshjohnson: I'm confused as to why this file is needed? 🤔

tinovyatkin: We use to check final (transpiled) output to prevent errors like non-ES5 complaint code or use of non-polyfilled objects.

PR #625, opened 2019-08-09
Made IE11 scrollbar check more precise (fixes #621)

sebastianhaberey:
Made IE11 scrollbar check more precise (fixes #621)

PR #710, opened 2019-10-27
Made IE11 scrollbar check more precise (updated) (fixes #621)

### LINKED ISSUE #621: IE11: Dropdown re-opens after selecting item
Happens randomly but frequently. Easy to replicate. Recorded directly from https://joshuajohnson.co.uk/Choices/:

test

Where should we start looking for the cause? Any pointers? Regards,

Sebastian

sebastianhaberey:
Updated #625 to match current master.

This is just the fix in the source code. To ensure effortless merging, I didn't rebuild the assets here.

### DIFF
--- a/src/scripts/choices.js
+++ b/src/scripts/choices.js
@@ -1401,9 +1401,16 @@ class Choices {
 
   _onMouseDown(event) {
     const { target, shiftKey } = event;
-    // If we have our mouse down on the scrollbar and are on IE11...
+
+    // If the target element is inside the choices list and we are on IE11...
     if (this.choiceList.element.contains(target) && isIE11()) {
-      this._isScrollingOnIe = true;
+      // ... but the click was on the right side of the items...
+      const firstItem = this.choiceList.getChild(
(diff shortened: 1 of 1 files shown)


src/scripts/choices.js
```diff
@@ -1401,9 +1401,16 @@ class Choices {
...
     const { target, shiftKey } = event;
-    // If we have our mouse down on the scrollbar and are on IE11...
+
+    // If the target element is inside the choices list and we are on IE11...
     if (this.choiceList.element.contains(target) && isIE11()) {
-      this._isScrollingOnIe = true;
+      // ... but the click was on the right side of the items...
+      const firstItem = this.choiceList.getChild(
+        `.${this.config.classNames.item}`,
+      );
```
tinovyatkin: @sebastianhaberey  This library support RTL, so, we need to check here `dir` and use either left or right

jshjohnson:
This has been superseded by https://github.com/jshjohnson/Choices/pull/740 :) thanks

Issue #689, opened 2019-10-24
RTL css is not working properly for selected value in dropdown.


sriram275:
With your RTL css everything is working except selected value is not working which is still positioning in left.  So is there way to not include RTL css so we will conver by using RTLcss convertor.
![image](https://user-images.githubusercontent.com/7270651/67483119-b850df80-f682-11e9-9f28-3597021e61d8.png)

Thanks.

tinovyatkin:
@sriram275 below is screenshot from [demo page](https://joshuajohnson.co.uk/Choices/#choices-single-rtl) and it shows everything correctly (the only problem in current version is scrollbar position):
![Screenshot 2019-10-27 at 15 12 59](https://user-images.githubusercontent.com/5350898/67639925-571c4c80-f8cc-11e9-905f-002ad2ea927e.png)
Are you applying some extra CSS?
Did you set `dir=rtl` attribute on original SELECT element?

tinovyatkin:
@sriram275 Could you check whether this problem is fixed for you in recently released Choices 8?

PR #711, opened 2019-10-28
[FIX] don't fallback to `ltr`

tinovyatkin:
Choices currently always explicitly sets `dir` attribute either to value specified by passed element or `ltr`. That's not correct - [W3C best practices for this case](https://www.w3.org/International/questions/qa-html-dir) directly stating that `dir` attribute should only be used on element if that particular element is _changing_ direction from the documents' one.
This PR changes behaviour to the recommended one.
That's also solves #689 and similar use cases.

### DIFF
--- a/src/scripts/choices.js
+++ b/src/scripts/choices.js
@@ -126,7 +126,23 @@ class Choices {
     this._wasTap = true;
     this._placeholderValue = this._generatePlaceholderValue();
     this._baseId = generateId(this.passedElement.element, 'choices-');
-    this._direction = this.passedElement.element.getAttribute('dir') || 'ltr';
+    /**
+     * setting direction in cases where it's explicitly set on passedElement
+     * or when calculated direction is different from the document
+     * @type {HTMLElement['dir']}
+     */
+    this._direction = this.passedElement.element.dir;
+    if (!this._direction) {
+      const { direction: elementDirection } = window.getComputedStyle(
+        this.passedElement.element,
+      );
+      const { direction: documentDirection } = window.getComputedStyle(
+        document.documentElement,
+      );
+      if (elementDirection !== documentDirection) {
+        this._direction = elementDirection;
+      }
+    }
     this._idNames = {
       itemChoice: 'item-choice',
     };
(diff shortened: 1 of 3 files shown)


src/scripts/choices.js
```diff
@@ -126,7 +126,18 @@ class Choices {
     this._wasTap = true;
     this._placeholderValue = this._generatePlaceholderValue();
     this._baseId = generateId(this.passedElement.element, 'choices-');
-    this._direction = this.passedElement.element.getAttribute('dir') || 'ltr';
+    /**
+     * setting direction is cases where it's explicitly set on passedElement
```
```suggestion
     * setting direction in cases where it's explicitly set on passedElement
```

src/scripts/choices.js
```diff
@@ -126,7 +126,18 @@ class Choices {
...
+     * or when calculated direction is different from the document
+     * @type {HTMLElement['dir']}
+     */
+    this._direction =
+      this.passedElement.element.dir ||
+      (({ direction }) =>
+        direction !==
+        window.getComputedStyle(document.documentElement).direction
+          ? direction
+          : '')(window.getComputedStyle(this.passedElement.element));
```
jshjohnson: This is not particularly easy to read - I think this could abstracted into it's own function or at least improved by using an if/else block

tinovyatkin: C'mon, that lovely use of Javascript closure and ternary 😄 I think moving this functionality into an external one line function will make it even harder to understand, as you will need to check that function code. And this comparison have no use anywhere else in the library.
if/else you mean like this?

```js
this._direction = this.passedElement.element.dir;
if (!this._direction) {
  const { direction } = window.getComputedStyle(this.passedElement.element);
  if (window.getComputerStyle(document.documentElement).direction !== direction)
    this._direction = direction;
}
```
Do you prefer that version?

jshjohnson: This would be better:

```js
this._direction = this.passedElement.element.dir;

if (!this._direction) {
  const { direction: elementDirection } = window.getComputedStyle(this.passedElement.element);
  const { direction: documentDirection }  = window.getComputerStyle(document.documentElement);
  
  if (elementDirection !== documentDirection) {
    this._direction = direction; 
  }
}
```

If statements should always use brackets unless it's a ternary

jshjohnson: (i appreciate this lib is already inconsistent in that style but it's where i'd like to end up)

jshjohnson: actually i'm gunna open a pr to enforce this styling 👍

PR #718, opened 2019-10-30
use last-of-type

tinovyatkin:
## Types of changes

## Checklist:

### DIFF
--- a/config/jsdom.js
+++ b/config/jsdom.js
@@ -42,6 +42,7 @@ global.HTMLSelectElement = window.HTMLSelectElement;
 global.HTMLInputElement = window.HTMLInputElement;
 global.DocumentFragment = window.DocumentFragment;
 global.requestAnimationFrame = window.requestAnimationFrame;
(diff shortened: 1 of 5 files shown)


justin-duncan:
Downloaded your fork and built it, but it doesn't solve what my issue was that it doesn't flip based on current view-port, now it seems like the default is always up, am I missing something?

tinovyatkin:
@justin-duncan you are right, sorry, seems like need to remove _more_ code here.
Is that looks right now?

Screenshot 2019-10-30 at 09 02 55

Screenshot 2019-10-30 at 09 02 43

justin-duncan:
Yeah, that's much better if its more efficient, appreciated @tinovyatkin.

tinovyatkin:
@justin-duncan it's more efficient at least due to fact that it [doesn't causes extra reflow(s)](https://gist.github.com/paulirish/5d52fb081b3570c81e3a) here

PR #721, opened 2019-10-30
fix async setChoices flow

twawszczak:
## Description
In async `setChoices` variant, wrap `_handleLoadingState(false)` into `requestAnimationFrame`. (variant with loadingState => true is already wrapped)

## Motivation and Context
Flow:
-> `requestAnimationFrame` -> `handleLoadingState(true)` -> insert placeholder, if needed
-> get data asynchronously
-> `handleLoadingState(false)` -> change placeholder content

So, there is a possibility, that `handleLoadingState(false)` will occur faster then `handleLoadingState(true)` (when async data request comes back very fast) So we trying to update placeholder content, when placeholder no initialized yet. (error, and hanging loading state after operation)  

## How Has This Been Tested?
Manually in browser.

## Types of changes

## Checklist:

### DIFF
--- a/src/scripts/choices.js
+++ b/src/scripts/choices.js
@@ -558,7 +558,9 @@ class Choices {
       }
 
       // it's a choices fetcher
-      requestAnimationFrame(() => this._handleLoadingState(true));
+      ({ then: requestAnimationFrame.bind(window) }.then(() =>
+        this._handleLoadingState(true),
+      ));
       const fetcher = choicesArrayOrFetcher(this);
       if (typeof fetcher === 'object' && typeof fetcher.then === 'function') {
         // that's a promise
@@ -569,6 +571,7 @@ class Choices {
               console.error(err);
             }
           })
+          .then(() => ({ then: requestAnimationFrame.bind(window) }))
           .then(() => this._handleLoadingState(false))
           .then(() => this);
       }


src/scripts/choices.js
```diff
@@ -569,7 +569,9 @@ class Choices {
               console.error(err);
             }
           })
-          .then(() => this._handleLoadingState(false))
+          .then(() =>
+            requestAnimationFrame(() => this._handleLoadingState(false)),
+          )
```
tinovyatkin: It was wrapped in `requestAnimationFrame` in original `.ajax` version. @jshjohnson do you remember why?

tinovyatkin: I'm not a fan of mixing promises and callbacks, as my head stops understand flow. Can we just return PromiseLike here if we must wait for animation frame?
```js
  .then(() => ({ then: next => requestAnimationFrame(next) }))
  .then(() => this._handleLoadingState(false));
```

twawszczak: @tinovyatkin For me no difference, it's not mine codebase, the only thing I need is to make it working properly. :)

twawszczak:
@tinovyatkin I have adjusted code to your suggestions. No problem to perform your own solution, but I will be thankful for solving that issue. :)

And BTW - IMO #727 Is absolutely related to that.

src/scripts/choices.js
```diff
@@ -558,7 +558,9 @@ class Choices {
       }
 
       // it's a choices fetcher
-      requestAnimationFrame(() => this._handleLoadingState(true));
+      ({ then: requestAnimationFrame.bind(window) }.then(() =>
+        this._handleLoadingState(true),
+      ));
```
tinovyatkin: This is not looking pretty. I think the whole problem here is not to wait for `_handleLoadingState` callback return. So, whole fragment can be rewritten like this:

```js
      // it's a choices fetcher
      const fetcher = choicesArrayOrFetcher(this);
      if (typeof Promise === 'function' && fetcher instanceof Promise) {
        // that's a promise
        // eslint-disable-next-line compat/compat
        return new Promise(requestAnimationFrame.bind(window))
          .then(() => this._handleLoadingState(true))
          .then(() => fetcher)
          .then(data => this.setChoices(data, value, label, replaceChoices))
          .catch(err => {
            if (!this.config.silent) {
              console.error(err);
            }
          })
          .then(() => this._handleLoadingState(false))
          .then(() => this);
      }
```

jshjohnson: I think trying to make `requestAnimationFrame` behave like a promise is just going to confuse people to be honest.

tinovyatkin: @jshjohnson do you mean to write it completely instead?
```js
new Promise(resolve => requestAnimateFrame(resolve))
```
it's just transpiles to an even longer version, so, `bind` was handy here... _Confused_ developer will not touch this line - that what we want, doesn't it?

jshjohnson: `new Promise(resolve => requestAnimateFrame(resolve))` is much more readable to `new Promise(requestAnimationFrame.bind(window)` in my opinion

twawszczak: @tinovyatkin @jshjohnson  I was trying with `new Promise()` but your prettier does not accept that. Please, implement it the best way you think.

PR #723, opened 2019-10-30
Fix lint

tinovyatkin:
Prevents lint errors for cases like https://github.com/jshjohnson/Choices/runs/281570154

Also, it seems that since Cypress 3.5 it tolerates lack of record key, so, on PR originated by @jshjohnson we should start seeing beautiful Cypress decorations.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -9,6 +9,7 @@ on:
       - '.babelrc'
       - 'webpack.config.*'
       - 'public/test/**'
+      - 'cypress/**'
       - '.github/workflows/cypress.yml'
 
 jobs:
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -1,6 +1,15 @@
 name: Cypress
 
-on: [pull_request]
+on:
+  pull_request:
+    paths:
+      - 'src/**'
+      - 'package-lock.json'
+      - '.browserslistrc'
+      - '.babelrc'
+      - 'webpack.config.*'
+      - 'public/test/**'
+      - '.github/workflows/cypress.yml'
 
 jobs:
   test-e2e:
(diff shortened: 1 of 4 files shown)


.github/workflows/cypress.yml
```diff
@@ -20,13 +30,13 @@ jobs:
...
+        run: npx run-p --race start cypress:ci
         env:
           CI: true
+          TERM: xterm-256color
           CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
           DEBUG: commit-info,cypress:server:record
           # https://docs.cypress.io/guides/guides/continuous-integration.html#Environment-variables
           COMMIT_INFO_BRANCH: ${{ github.head_ref }}
-          COMMIT_INFO_AUTHOR: ${{ github.event.sender.login }}
-          COMMIT_INFO_SHA: ${{ github.event.after }}
```
jshjohnson: What was the reason for removing these environment variables? Since merging i'm getting errors like the following from cypress: https://github.com/jshjohnson/Choices/runs/283664081#step:5:0 :)

tinovyatkin: @jshjohnson I've added them just to help `commitinfo`, but it was able to find the same information in the previous run. You may revert that, it doesn't hurt to have them here.

tinovyatkin: But errors seems unrelated

jshjohnson: yeah i think i know where the problem is - we're not passing a value to `--group` in the `cypress:ci` script

tinovyatkin: @jshjohnson github.event.after will not be available on `push` to master, so, if you would like to reuse the same action you may better remove COMMIT_INFO_SHA completely here

PR #729, opened 2019-10-31
Add documentation label + headings

jshjohnson:


tinovyatkin:
@jshjohnson may be 'test' needs a label too? I think for many people test coverage and extensiveness is important while selecting a library...

jshjohnson:
I would say that goes under `housekeeping`

PR #731, opened 2019-11-01
Fix #727

### LINKED ISSUE #727: Placeholder not hiding when loading data
Example: 
Screenshot 2019-10-31 at 12 00 59

Input types affected: `select-one` 

Markup:
```html
        <div data-test-hook="remote-data">
          <label for="choices-remote-data">Remote data</label>
          <select
            class="form-control"
            name="choices-remote-data"
            id="choices-remote-data"
          >
            <option value="">This is a placeholder</option>
          </select>
        </div>
```

JS:
```js
        new Choices('#choices-remote-data', {
          shouldSort: false,
        }).setChoices(async () => {
          const res = await fetch('/data');
          return res.json();
        });
```

jshjohnson:
## Description

Fixes the issue detailed in #727. 

This removes the ability to pass a placeholder via config to select elements - the implementation was confusing and semantically it is better to pass an `option` element. It also removes placeholder choices from the choices list for select multiple elements. 

## Types of changes

## Checklist


src/scripts/choices.js
```diff
@@ -846,7 +848,7 @@ class Choices {
     let choiceLimit = rendererableChoices.length;
 
     // Prepend placeholeder
```
```suggestion
    // Prepend placeholder
```

tinovyatkin:
@jshjohnson IE is green :tada: 
download the puppeteer artifact, inspect and replace it as part of this PR

jshjohnson:
> @jshjohnson IE is green 🎉
> download the puppeteer artifact, inspect and replace it as part of this PR

🎉 if you can think of a better solution to https://github.com/jshjohnson/Choices/pull/731/commits/59840d14c694ce93f33d978124eab321e8705763 let me know :D i've run out of free cypress recordings for the month

tinovyatkin:
> > @jshjohnson IE is green 🎉
> > download the puppeteer artifact, inspect and replace it as part of this PR
> 
> 🎉 if you can think of a better solution to [59840d1](https://github.com/jshjohnson/Choices/commit/59840d14c694ce93f33d978124eab321e8705763) let me know :D i've run out of free cypress recordings for the month

Sure: did you apply to https://www.cypress.io/oss-plan/ ?

tinovyatkin:
@jshjohnson So, we are gonna have next major soon? I have a couple of breaking changes in mind...

jshjohnson:
> @jshjohnson So, we are gonna have next major soon? I have a couple of breaking changes in mind...

Unfortunately yes. Feel free!

jshjohnson:
> Sure: did you apply to https://www.cypress.io/oss-plan/ ?

I've applied now 🎈

Fedik:
Sorry guys but the fix made the things even more confusing.

What if I want to be able to select an empty options, in multiple list? 

```
<select multiple>
  <option value="11">Option 1</option>
  <option value="">Option 2</option>
</select>
```
After the fix `<option value="">Option 2</option>` will be considered as placeholder that is wrong.

An additionally, when I build the same with AJAX.

jshjohnson:
> What if I want to be able to select an empty options, in multiple list?

Why would you want to do that? That's the equivalent of selecting nothing 🤔

Fedik:
> That's the equivalent of selecting nothing

to select a "nothing" ;)

but it not a nothing: 
```
<select multiple>
  <option value="11" selected>Option 1</option>
  <option value="" selected>Option 2</option>
</select>
```
will give you: `["11", ""]` that is a "something"

PR #734, opened 2019-11-02
Add missing type definitions + rename sortFn

jshjohnson:
## Description

- Update types file
- Utilised type defs throughout
- Minor refactoring
- Renames `sortFn` to `sorter` to be consistent with other functions passed via config
- Renames `itemComparer` to `valueComparer` to be clearer about what the function actually does

## Types of changes

## Checklist


tinovyatkin:
@jshjohnson can we rename `sortFn` to `sortProvider`? My breaking PR is to introduce `searchProvider` and make it easy to build autocomplete on top of Choices

tinovyatkin:
That's also can be done in a way of deprecation - I mean, undocument old settings now, but still support it until next major.

tinovyatkin:
BTW, my `searchProvider` will be support promises (similar to setChoices), not sure if that makes sense for sorting, but...

jshjohnson:
> BTW, my `searchProvider` will be supporting promises (similar to `setChoices` approach), not sure if that makes sense for sorting, but...

Hmm, hard to say without seeing your implementation. Do you mind if I keep this as is for the time being and if we think we need to update it after your search PR, we can?

PR #740, opened 2019-11-05
improve whole-page performance by scoping events handlers

tinovyatkin:
## Description

This is a further improvement over #703 and #696. We have `mousedown` event handler in capture mode for `document.documentElement`, while inside of it immediately returning if element is not part of outerContainer. So, that logically means we should initially set our target element properly. 

Similar logic was applied to `keydown` and `mouseover` handlers.

Also fixed here invalid arguments and target for `removeEventListener`, introduces by #703 (sorry 🤷🏼‍♂️)

While we were here, reimplemented #710 in a little bit more performant way and with support for LTR/RLT directions.

Noticing, that `isIE11` util had being used only at mouse down handler and that result of this function call are always constant for the same browser, replacing it with a constant and using feature detection instead of userAgent sniffing.

As a result of the above refactorings removed two obsolete helpers from `utils`.

## Types of changes

## Checklist


src/scripts/choices.js
```diff
@@ -1572,47 +1573,55 @@ class Choices {
...
     }
 
-    if (
-      !this.containerOuter.element.contains(target) ||
-      target === this.input.element
-    ) {
+    // If we have our mouse down on the scrollbar and are on IE11...
+    if (IS_IE11 && this.choiceList.element.contains(target)) {
+      // check if click was on a scrollbar area
+      const firstChoice = /** @type {HTMLElement} */ (this.choiceList.element
```
jshjohnson: Can we put the comment above the variable please :) then you don't need to do the braces:

```js
/** @type {HTMLElement} */
const firstChoice =  this.choiceList.element.firstElementChild;
```

jshjohnson: Like we do here: https://github.com/jshjohnson/Choices/blob/master/src/scripts/choices.js#L53-L54

tinovyatkin: it's better to give a type to `choiceList.element`, so, VSCode will pick it up here. Do you want me to do that as part of this PR?

PR #744, opened 2019-11-07
Always hide remove button if no value is selected

### LINKED ISSUE #684: Only show clear results x when there's input to clear
As a usability enhancement, I propose only having the clear results "x" button for single selects only become visible/accessible once there's an option chosen to clear.

This should help reduce confusion/clutter around the x when no option is selected.

bronzehedwick:
## Description

In some cases, the remove item X button is not hidden when there is no
selection made from the dropdown. This is not as usable, since there's
no item to clear, hence an effective noop.

The condition I saw seeing this issue arise is in single selects with
a placeholder value, a selection has been made, and then cleared: the
button HTML remains.

To fix, add some styles to hide the button if the select doesn't have a
value selected, in this case keying off of it's parents `data-value`
property.

fixes #684

## Types of changes

## Checklist


bronzehedwick:
@jshjohnson Not sure why the selenium tests are failing, the error messages are pretty vague. Do you have any insight here? Thanks.

jshjohnson:
> @jshjohnson Not sure why the selenium tests are failing, the error messages are pretty vague. Do you have any insight here? Thanks.

I wouldn't worry about that - Github has made some changes to the macOS-latest image which is causing problems with CI 👍

jshjohnson:
Thanks @bronzehedwick  👍

Issue #622, opened 2019-08-09
Unable to run tests on Windows


sebastianhaberey:
There are a few *nix-only commands in package.json that prevent test execution on windows. I fixed these and added a missing dependency in travis.yml to get them running again. See [pull request](https://github.com/jshjohnson/Choices/pull/623).

PR #748, opened 2019-11-08
Upgrade Cypress to 3.6 and use cache in Github Actions

tinovyatkin:
This PR upgrades Cypress to the current version (3.6).
Using new caching functionality caches Cypress cache what resulting in running time of ~2 mins, instead of current >3 min.
It also tunes the action in general, fixing some small errors.
Also removing unneeded prefixes in `package.json`, superseding stale #622


.github/workflows/e2e-tests.yml
```diff
@@ -10,45 +10,58 @@ on:
...
-          COMMIT_INFO_AUTHOR: ${{ github.event.sender.login }}
-          COMMIT_INFO_SHA: ${{ github.event.after }}
-
-      # if we have ran out of free Cypress recordings, run Cypress with recording switched off
-      - name: run Cypress (without recording)
-        if: failure()
-        run: npx run-p --race start cypress:run
+      - name: run Cypress (with or without recording)
+        # if we have ran out of free Cypress recordings, run Cypress with recording switched off
+        run: npx run-p --race start cypress:ci || npx run-p --race start cypress:run
```
jshjohnson: Clever!

jshjohnson:
Thanks! :)

PR #749, opened 2019-11-09
Update choices.scss. Fixed a bug with the height of the source element.

dios-oleg:
## Description
Fixed a bug with the height of the source element. Added property "box-sizing: border-box;" 

Before: https://yadi.sk/i/tsTioPhHB3QNgg
After: https://yadi.sk/i/R1jkMhqK_jDKLQ

## Types of changes
Fixed CSS.

## Checklist

### DIFF
--- a/src/styles/choices.scss
+++ b/src/styles/choices.scss
@@ -152,6 +152,7 @@ $choices-icon-cross-inverse: url(data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMjEiI
   font-size: $choices-font-size-md;
   min-height: 44px;
   overflow: hidden;
+  box-sizing: border-box;
   .is-focused &,
   .is-open & {
     border-color: darken($choices-keyline-color, 15%);


src/styles/choices.scss
```diff
@@ -152,6 +152,7 @@ $choices-icon-cross-inverse: url(data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMjEiI
   font-size: $choices-font-size-md;
   min-height: 44px;
   overflow: hidden;
+  box-sizing: border-box;
```
jshjohnson: Instead of applying this to specific selectors, could we not just apply to everything within the .choices class?

Something like:

```css
.choices * {
  box-sizing: border-box;``
}
```

dios-oleg: Yes, we can apply to everything selectors, but we will lose a little performance.
I noticed that this is necessary for two classes.

tinovyatkin: Universal selectors are slow and box-sizing is inheriting to children, so, agree with @dios-oleg that those styles better be applied when needed.

PR #801, opened 2019-12-26
Reduce bundle size

jshjohnson:


stof:
If you want to reduce the bundle size when using webpack as bundler, the main thing to do is to make sure that you run webpack in production mode (maybe disabling the minimizer if you want a non-minified file) rather than in dev mode.
The dev mode trades bundle size in favor of compilation speed by disabling all optimization. And while minimization can (and should) be applied when importing Choices.js in your project, other optimization like the module concatenation cannot (as they are not modules anymore). In particular, the module concatenation optimization has a big impact on the output size (especially after minification, as its output minifies better by allowing mangling a lot more name) and also benefits performance (by avoiding to have one closure per original module).

Note that to benefit from this optimization, you must configure Typescript to output ES modules rather than CommonJS ones, because webpack applies that optimization only on ES modules.

mtriff:
Closing this for now...

Issue #806, opened 2020-01-13
Remote search


sharkooon:
Using Choices for a while and now I want to implement remote search. 
The list of users is over 100k, so external load and search is needed.

But it loses focus on every keystroke when typing.
Please see jsfiddle example: https://jsfiddle.net/8zn132go/9/

```
var config = {
  'searchResultLimit': 15,
  'searchChoices': false
};
var select = document.getElementById('users');
var choice = new window.Choices(select, config);
select.addEventListener('search', function(event) {
    if (event.detail.value) {
        choice.setChoices(function(callback) {
            return fetch(
                // below api is only used for this example. it doesn't support query filtering, but its returns some data. 
                'https://randomuser.me/api/?results=20&q=' + event.detail.value
            )
            .then(function(response) {
                return response.json();
            })
            .then(function(data) {
                return data.results.map(function(user) {
                    return { value: user.login.uuid, label: user.name.first };
                });
            });
        })
    }
});
```

daianacambruzzi:
I have the same problem :/  @sharkooon did you fix the issue with any alternative?

sharkooon:
not yet, waiting for reply.

DoctorOak:
Hitting the exact same problem and this is the specific functionality I need the plugin for...

I tried putting a timeout throttle around firing the request in my code but it just caused even more problems. I'm assuming a lack of any kind of throttle within the core code itself might be the problem but I've not really got the time to look into it. 

Until this can be solved, I'm going to have to revert to Select2, which is a pity.

jshjohnson:
This looks like a bug - will look into this 👍

sharkooon:
any updates? otherwise i have to implement Selectize

bjornpost:
@sharkooon did you find a solution for this/did you investigate further? Was your code example working in a previous version of Choices?

sharkooon:
Don't know if it worked in previous versions.. and did not found a solution for it, i used Selectize now.

Raidri:
@jshjohnson  I think just set 
this.input.element.focus();
at line 356 in the enable function should solve the problem. In my test the focus still on the input element

medeirosrafael:
I had same problem.. 
I solved calling .focus() on my input, like @Raidri suggested..

My code:

```  
 myInput.addEventListener('search', (event) => {
      myChoice.setChoices(async () => {
        return this.doSearch(event.detail.value);
      })
      .then(() => {
[...]

PR #827, opened 2020-02-25
fix bug #806

Raidri:
## Description
Fixed the bug #806.
Added to the input element the focus() function. It solves the problem that the focus get lost after a remote search.

## Types of changes

## Checklist


Xon:
I plan to investigate more as it looks like how the input field is being disabled while data is being loaded is just disruptive and probably needs a re-think

ritwikBiswas007:
`<select id="users"></select>`

```
const select = document.getElementById('users');

const choice = new Choices(select, {
    searchChoices: false,
    searchResultLimit: 15,
    shouldSort: false,
    removeItemButton: false,
});

// Debounce function to limit how often we call API
function debounce(func, delay) {
    let timeout;
    return function (...args) {
        clearTimeout(timeout);
        timeout = setTimeout(() => func.apply(this, args), delay);
    };
}

// Debounced search handler
const handleSearch = debounce(async function (event) {
    const searchTerm = event.detail.value;

    if (!searchTerm) return;

    const input = document.querySelector('.choices__input');

    // Show loading text
    choice.clearChoices();
    choice.setChoices([{ value: '', label: 'Loading...', disabled: true }], 'value', 'label', false);

    try {
        const response = await fetch(`https://randomuser.me/api/?results=10`);
        const data = await response.json();

        const newChoices = data.results.map(user => ({
            value: user.login.uuid,
            label: `${user.name.first} ${user.name.last}`,
        }));

        choice.setChoices(newChoices, 'value', 'label', true);

        // Refocus the input
        setTimeout(() => {
            const newInput = document.querySelector('.choices__input');
            if (newInput) {
                newInput.focus();
                // Restore cursor position
                const val = newInput.value;
                newInput.value = '';
                newInput.value = val;
            }
        }, 30);
    } catch (error) {
        console.error('API error:', error);
        choice.setChoices([{ value: '', label: 'Error loading users', disabled: true }], 'value', 'label', false);
    }
}, 300);

// Add listener
select.addEventListener('search', handleSearch);
```

This code fixed the issue

PR #861, opened 2020-04-24
Feat/allow to be imported on server

adammockor:
This should add possibility to import this library in SSR environments.

### DIFF
--- a/src/scripts/choices.ts
+++ b/src/scripts/choices.ts
@@ -40,6 +40,7 @@ import {
   generateId,
   existsInArray,
   diff,
+  canUseDom,
 } from './lib/utils';
 import {
   Options,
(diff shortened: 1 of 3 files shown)


mtriff:
This seems reasonable. Can you rebase on the latest version of `master` and add an e2e test for server side rendering? [This](https://glebbahmutov.com/blog/ssr-e2e/) seems like a possible approach to that.

Xon:
I can't see any point updating `strToEl` when all the template rendering callbacks would also need changing as they use `document.createElement` as well.

The `canUseDom` needed to be used in in another two spots. With rollup building multiple copies (specifically the cjs/mjs export formats), I think this also resolves the `globalObject` object. 

@adammockor could you have a look at https://github.com/Xon/Choices.js/commit/0d5d93b419a13b5812093c9904eaf10adc8448f6 and let me know if this is working in the given SSR environment but I think additional tests would be required before this could be merged.

Xon:
This ended up being useful for running choices.js in tests in v11.0.0, but required some tinkering to get tree-shaking working sanely

PR #863, opened 2020-05-09
Allow overriding language string for removing an item

wilsonge:
## Description
Allows overriding the language string per the TODO in the code. Required for the multilingual setup we're using this for in @joomla

## Types of changes

I'm not 100% on whether this will be a breaking change for people overriding the item template (with the introduction of the new parameter). I think it will be fine but I don't have an implementation of that to hand to test with

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ Or include Choices directly:
     loadingText: 'Loading...',
     noResultsText: 'No results found',
     noChoicesText: 'No choices to choose from',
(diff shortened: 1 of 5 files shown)


mtriff:
This looks great! Can you rebase on the latest version? It would also be nice to have a test for this.

Xon:
Thanks! Implemented as part of https://github.com/Choices-js/Choices/pull/1166

PR #882, opened 2020-07-24
correct the fuse.js option in readme

paulovieira:
The option that is given as example doesn't seem to be valid (according to https://fusejs.io/api/options.html).

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ Or include Choices directly:
     // Choices uses the great Fuse library for searching. You
     // can find more options here: https://github.com/krisk/Fuse#options
     fuseOptions: {
-      include: 'score'
+      includeScore: true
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -170,7 +170,7 @@ Or include Choices directly:
     // Choices uses the great Fuse library for searching. You
     // can find more options here: https://github.com/krisk/Fuse#options
```
shawnbot, line 171: I also noticed that the URL for Fuse options is wrong; the new one is:

```suggestion
    // more options here: https://fusejs.io/api/options.html
```

mtriff:
This library is using an old version of Fuse.js and the options listed on the new site are not correct. We'll update this when we upgrade to Fuse v6.

PR #945, opened 2021-04-16
Trigger changeQuery event when user inputs something into the search field

### LINKED ISSUE #944: Trigger an event as soon as the user inputs something in the search input field
**Is your feature request related to a problem? Please describe.**
We would like to hook into the moment a user types something in the search field, but before the search actually happens (so the `search` event is not an option).

**Describe the solution you'd like**
An event, such as one called `changeQuery`, gets sent as soon as the user types into the field.

SanderKleykens:
## Description
This closes #944 by triggering a `changeQuery` event as soon as the user inputs something into the search field.

## Types of changes

## Checklist


mtriff:
This looks like a great addition, but right now it won't function as intended. A `changeQuery` event will be fired every time a key is pressed, even if it doesn't actually change the query (i.e. arrow keys, delete/backspace keys when the query box is empty, etc.). I think you want to hook into the `input` event instead. We aren't doing that already, so you'll need to add an `_onInput` method to `Choices`.

Please also rebase on the latest version of the `master` branch.

src/scripts/choices.test.ts
```diff
@@ -2161,6 +2161,24 @@ describe('choices', () => {
...
+          'changeQuery',
+          event => {
+            expect(event.detail).to.eql({
+              value: 'foobar',
+            });
+            done();
+          },
+          false,
+        );
+
```
mtriff, line 2177: Let's also add a check to ensure that `changeQuery` is fired before the `search` event. You can see an example of firing the `search` event [here](https://github.com/Choices-js/Choices/commit/9605f367f9021324f5d03023005be1086ace9954#diff-12fe009a4650f4b3956664b92225c8a78cd2fc3c9fd22667a0161131062c130dR1739-R1766).

PR #968, opened 2021-09-22
fix: xss vulnerability in creating choice labels

hrahman-miovision:
## Description

This addresses a XSS vulnerability when creating labels for choices. For example, if we set the label property of a choice to the HTML string `` like so:
```
var singleNoSearch = new Choices('#choices-single-no-search', {
          searchEnabled: false,
          removeItemButton: true,
          choices: [
            { value: 'One', label: '<img src=x onerror=alert()>' },
            { value: 'Two', label: 'Label Two', disabled: true },
            { value: 'Three', label: 'Label Three' },
          ],
        });
```
It will execute the JavaScript function defined in the `onerror` attribute since the `src` does not exist. An attacker can exploit this.

The issue was that when creating the choice / placeholder elements, it would set the `innerHTML` content of the element. Instead, since these are just string labels, we should be setting the content as a string using the `innerText` property.

Choices executing JavaScript functions (`alert()`) defined in a label string.
choices-js-xss

## Types of changes

## Checklist

### DIFF
--- a/cypress/integration/text.spec.ts
+++ b/cypress/integration/text.spec.ts
@@ -45,7 +45,7 @@ describe('Choices - text element', () => {
               .should($dropdown => {
                 const dropdownText = $dropdown.text().trim();
                 expect(dropdownText).to.equal(
-                  `Press Enter to add "${textInput}"`,
+                  `Press Enter to add <b>"${textInput}"</b>`,
                 );
               });
           });
--- a/src/scripts/templates.ts
+++ b/src/scripts/templates.ts
@@ -296,7 +296,7 @@ const templates = {
       noResults,
       noChoices,
     }: Pick<ClassNames, 'item' | 'itemChoice' | 'noResults' | 'noChoices'>,
-    innerHTML: string,
+    innerText: string,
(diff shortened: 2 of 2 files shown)


mtriff:
Thanks for the contribution, Hasib. However, this would be a breaking change for anyone using HTML in their choices (for icons, font styling, etc.) I understand this is a XSS vulnerability, but only if you are loading your choices from an untrusted/third-party source. That said, I think it's a good idea to have a default safe option for users. I'd propose the following:

1. Add an `allowHTML` option. If that option is `true`, set `innerHTML`. If it's `false` set `innerText`.
2. For now, default to `allowHTML: true` so that we aren't introducing a breaking change.
3. If `allowHTML` is not explicitly set, log a deprecation warning.

In a future release we'll change the default to `allowHTML: false`.

What do you think? Would you be able to implement this change?

victiondev:
@mtriff I have created a new PR which completes these 3 things you have asked for. See #984

mtriff:
Thanks for flagging this @hrahman-miovision, closing in favour of #984

PR #976, opened 2021-12-13
Solve deprecated warning using / for division outside of calc()

hans2103:
## Description

This PR will solve the Deprecated Warning created by Dart Sass 

```
Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div(-$choices-button-offset, 2) or calc(-$choices-button-offset / 2)

More info and automated migrator: https://sass-lang.com/d/slash-div

    ╷
129 │     margin-right: -$choices-button-offset/2;
    │                   ^^^^^^^^^^^^^^^^^^^^^^^^^
    ╵
    node_modules/choices.js/src/styles/choices.scss 129:19                @import

```

## Types of changes

To solve the deprecated warning using `/` for division outside of calc() you can use math.div instead. 
BUT... does math.div have a widespread support?
Dunno... on the other hand.... the multiplier `*` does have a widespread support.

SO instead of replacing `/` by math.div I propose to replace it by `*`
There are only five places in the code where the division is used and the numbers are simple

- replace `/2` by `* .5`
- replace `/4` by `* .25`

## Checklist

There is no need to change the documentation

### DIFF
--- a/src/styles/base.scss
+++ b/src/styles/base.scss
@@ -65,7 +65,7 @@ h4,
 h5,
 h6 {
   margin-top: 0;
-  margin-bottom: $global-guttering * .5;
+  margin-bottom: $global-guttering * 0.5;
   font-weight: 400;
   line-height: 1.2;
 }
@@ -152,7 +152,7 @@ label + p {
 
 .logo {
   display: block;
-  margin-bottom: $global-guttering * .5;
+  margin-bottom: $global-guttering * 0.5;
 }
 
 .logo__img {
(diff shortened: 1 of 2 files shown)


mtriff:
Thanks for this! I agree your approach seems the most straightforward. Can you add the leading zeroes to `.5` and `.25`? We should add stylelint to automatically detect and fix that, but for now...

hans2103:
@mtriff leading zeroes have been added.

mtriff:
Looks great, thanks!

PR #984, opened 2021-12-23
feat: Introduce allowHTML option to allow people to disable injecting HTML into choices.

victiondev:
## Description

This is a continuation of #968 to address an XSS vulnerability when creating labels for choices.

Choices executing JavaScript functions (`alert()`) defined in a label string.
image

## Types of changes

## Checklist


mtriff:
Awesome! Thanks for taking this up. There are a few test and linting errors that need to be resolved. Can you also add Cypress tests for the three states of `allowHTML` (`true`, `false`, `undefined`)?

victiondev:
Should be done 👍Let me know if there are any other changes you would like to be done.

cypress/integration/text.spec.ts
```diff
@@ -355,11 +359,53 @@ describe('Choices - text element', () => {
       });
     });
 
+    describe('allow html', () => {
```
mtriff, line 362: These look good. Can you add these tests to the `select-multiple` and `select-one` specs as well?

mtriff (changes requested):
Almost there, still a couple of linting issues. It also looks like we have an issue with custom templates. You can see in the demo page that the custom template example is no longer rendering correctly (there's a unit test failure that should point you in the right direction too).

cypress/integration/text.spec.ts
victiondev: Resolved in [#3633c4a](https://github.com/Choices-js/Choices/pull/984/commits/3633c4ac0f0a8437ea4ffdc44de0b872c8e816c5) and [#c989be1](https://github.com/Choices-js/Choices/pull/984/commits/c989be14917d58c6f0969218289299d42bf69338)

src/scripts/templates.ts
```diff
@@ -4,14 +4,17 @@
...
 
 import { Choice } from './interfaces/choice';
-import { ClassNames } from './interfaces/class-names';
 import { Group } from './interfaces/group';
 import { Item } from './interfaces/item';
 import { PassedElementType } from './interfaces/passed-element-type';
 
+// Not the most ideal solution to requiring Options from ./interfaces/options and creating a circular dep
+// Fixing the circular would require merging the interfaces/templates or doing something like below.
+type Options = Record<'classNames' | 'allowHTML', any>;
```
mtriff, line 13: That makes sense. Let's rename this `TemplateOptions` to avoid confusion.

mtriff (changes requested):
This is excellent, one minor nitpick. Also, the `callbackOnCreateTemplates` advanced example in the README needs to be updated too. Then we are good to merge.

victiondev:
All done! 👍

mtriff (approved):

mtriff:
Excellent, thanks for this! I made some small changes to the wording around `allowHTML` and set `allowHTML: false` on our remote fetching examples (we should probably model good behaviour 😄).

victiondev:
Thanks for merging this, is there any chance this will be included in a release soon? Also, while I'm here I have a question regarding the typings because Im not able to import choices.js with typings through `import Choices from 'choices.js` or `import { Choices } from 'choices.js'`.

Seems like the typings file package.json points to isn't shipped with the module on npm so I can't make it work for me in a TS vue project.

mtriff:
I'd expect a release in the next week or two. There's at least one more breaking change I'd like to get merged first.

Thanks for flagging the types issue, looks like this was missed when the project was converted to TypeScript. It has been fixed on `master` (see #985). `import Choices from 'choices.js'` should work again.

victiondev:
Just had another go with the new version with the supposed typings fix - still no luck on my end.

Judging off of [this diff](https://github.com/Choices-js/Choices/commit/bf94386cc56a284c5e8408988e322e56d0b3e2f0#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R6), the npm module should ship with a public/types folder, which it doesn't seem to do.

image

mtriff:
Are you installing from the master branch? 
```
npm install Choices-js/Choices#master
```
There won't be a new version pushed to npm until the next release.

victiondev:
Yes, that's the exact command I ran to install it (I also uninstalled the release version beforehand)
```
    "choices.js": "github:Choices-js/Choices#master",
```

This is the line in my package.json, and the contents of the module are the same as above still.

victiondev:
I've created a PR which fixes this issue: #986

PR #1001, opened 2022-01-17
[BUGFIX] Correct evaluation of HTML custom properties and JSON support

brosua:
## Description

The data attribute was mispelled. Also I added a logic to parse JSON, if multiple custom attributes have to be added. 
It fixes #982.

## Types of changes

- [ ] Chore (tooling change or documentation change)
- [ ] Refactor (non-breaking change which maintains existing functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [x] My code follows the code style of this project.
- [ ] I have added new tests for the bug I fixed/the new feature I added.
- [ ] I have modified existing tests for the bug I fixed/the new feature I added.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1027, opened 2022-04-01
Fix custom properties for browsers

sofiyareverse:
Custom properties does not work correct for Chrome and Firefox Browsers. It returns 'undefined' and [object Object] from client side.

To reproduce it just add 'data-custom-properties' to your option and check it in Chrome or Firefox like that:
`<option value="1" data-custom-properties='<span>Primary</span>'>1</option>`

This small change fixes custom properties fields for browsers. Without it only Safari supports custom properties.

- [ x ] Bug fix (non-breaking change which fixes an issue)
- [ x  ] My code follows the code style of this project.

### DIFF
--- a/public/assets/scripts/choices.js
+++ b/public/assets/scripts/choices.js
@@ -399,7 +399,7 @@ function () {
           selected: !!option.selected,
           disabled: option.disabled || option.parentNode.disabled,
           placeholder: option.value === '' || option.hasAttribute('placeholder'),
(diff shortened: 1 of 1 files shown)


christiaan:
We've ran into this as well. This used to work in 9.0.1

mtriff:
I believe this was resolved by #1001 and should be fixed in the next release.

PR #1058, opened 2022-09-20
[BUGFIX] Check for constructor name of element on bootstrap

### LINKED ISSUE #1057: Cannot use Choices if rendered from iframe
**Describe the bug**
Disclaimer: To be honest, this is a kinda special setup.

I have a modal window whose content contains a `select` field that is used to initialize Choices.js. There are two scopes where this modal may be opened:

* in a simple standalone view
* in a more complex environment with an `iframe`, where the modal is rendered in the *top* frame

In the latter case, I get the error "Expected one of the following types text|select-one|select-multiple". The issue is that the `instanceof` check for the passed element to initialize Choices.js does not work in case of an inter-frame communication. In my tests it was safe to check `passedElement.constructor.name` instead.

Since the fix is rather easy I'll push a PR ASAP.

**To Reproduce**
1. In the iframe, have a callback that initializes Choices.js
2. Open a modal with the matching `select` element on the top frame
3. Execute the callback
4. See "Expected one of the following types text|select-one|select-multiple"

**Expected behavior**
I expect that Choices.js works in the scenarios mentioned above.

**Desktop (please complete the following information):**
 - OS: Fedora Linux 36
 - Browser Chrome
 - Version 105

andreaskienast:
## Description

The previous check using `instanceof` fails in case the Choices.js component is rendered in the `top` frame while the JavaScript actually using Choices.js is in an iframe.

The conditions are changed to check for `constructor.name` now, solving the issue.

Due to the rather messy constructor, some quirks like operational chaining and explicit type casts need to be used to make the TypeScript compiling process happy.

This fixes #1057.

## Types of changes

## Checklist


public/assets/scripts/choices.js
```diff
@@ -306,7 +306,7 @@ function () {
 
     var passedElement = typeof element === 'string' ? document.querySelector(element) : element;
 
-    if (!(passedElement instanceof HTMLInputElement || passedElement instanceof HTMLSelectElement)) {
+    if (!((passedElement === null || passedElement === void 0 ? void 0 : passedElement.constructor.name) === 'HTMLInputElement' || (passedElement === null || passedElement === void 0 ? void 0 : passedElement.constructor.name) === 'HTMLSelectElement')) {
```
mtriff, line 309: This change is more complex than what you've done elsewhere, does it need to be?

Assuming they can all be the same, let's move this functionality out to a function in `utils.ts`:
```javascript
export const isHTMLInputOrSelectElement = (element: HTMLElement): boolean => {
  const constructorName = element?.constructor.name;
  if (constructorName === 'HTMLInputElement' || constructorName === 'HTMLSelectElement') {
    return true;
  }
  return false;
}
```

public/assets/scripts/choices.js
```diff
@@ -4614,7 +4614,7 @@ var templates = {
     div.setAttribute('aria-expanded', 'false');
 
     if (labelId) {
-      div.setAttribute('aria-labeledby', labelId);
+      div.setAttribute('aria-labelledby', labelId);
```
mtriff, line 4617: Thanks for fixing this, can you also update it everywhere else?

Xon:
This can be more simply implemented using the syntax;
```ts
export const isHtmlInputElement = (e: Element): e is HTMLInputElement =>
  e.tagName === 'INPUT';
  ```
When this returns true, type script uses the `e is HTMLInputElement` to inform what the type if for a branch depending on it being true.

I'll implement this in v11.0.x branch I'm working on as that html element guard style is already used.

Issue #1068, opened 2022-11-08
Wrong character when using non-English (Farsi) keyboard


ousmorez:
**Describe the bug**
When we try to enter a non-English word for example Persian (Farsi), this happens:
While the keyboard is set on Farsi language, open a select box and then select one of the options. Then at that moment try to write something, as you see in the GIF below, the first letter is English and then the second letter and the rest of them is Persian (Farsi).
![choicejs-bug](https://user-images.githubusercontent.com/50699824/200555596-acb38e19-b706-4388-b5a1-3a7caaccecd3.gif)

Also, if you try to enter a number (numbers in the image below), the first character will be an English letter, then the second character and the rest will be numbers.

![71Va9DgRgwL _SL1500_](https://user-images.githubusercontent.com/50699824/200556121-a0b9421c-dab1-4a6a-9f55-a294c77133bd.jpg)

**To Reproduce**
1. Set your keyboard on Persian (Farsi)
See how to add languages to keyboard on windows 10/11 :
https://support.microsoft.com/en-us/windows/manage-the-input-and-display-language-settings-in-windows-12a10cb4-8626-9b77-0ccb-5013e0c7c7a2#:~:text=Select%20the%20Start%20button%2C%20then,keyboard%20you%20want%20to%20add.
2. Go to [Choicejs](https://choices-js.github.io/Choices/)
3. Click on 'Right-to-left'
4. Select 'Choice1'
5. Type something
6. First letter is English and the rest of the letters is Persian (Farsi)

**Expected behavior**
I expected to see only Farsi (Persian) characters on the screen.

**Screenshots**
![choicejs-bug](https://user-images.githubusercontent.com/50699824/200555596-acb38e19-b706-4388-b5a1-3a7caaccecd3.gif)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: Latest

I need to use Choicejs asap, if someone tell me a solution I can fix it in my project, I use it in formiojs and I can fix it externally.
If I find the solution myself, I'll post a merge request here, but I need help.

PR #1072, opened 2022-11-17
[BUGFIX] Fix support for non-Latin characters

ousmorez:
## Description

I've added some more characters to the dropdown opener in the regex, so that the dropdown opens when the following keys are pressed on the keyboard:
MacTigr-thumb-top
Because if those characters are available to type, they can open the drop-down menu after pressing.
Also correct first character when trying to type numbers or non-English.

it solves following bug:

https://github.com/Choices-js/Choices/issues/1068

## Types of changes

## Checklist


src/scripts/choices.ts
```diff
@@ -1438,8 +1438,8 @@ class Choices implements Choices {
     const hasActiveDropdown = this.dropdown.isActive;
     const hasItems = this.itemList.hasChildren();
     const keyString = String.fromCharCode(keyCode);
-    const wasAlphaNumericChar = /[a-zA-Z0-9-_ ]/.test(keyString);
-
+    const wasAlphaNumericChar = /[a-zA-Z0-9-_ |À|¿|¼|º|Þ|Ü|Û|Ý|¿|¾|`]/.test(keyString);
```
mtriff, line 1441: Instead of continually adding characters to this check, let's instead check to see if user has entered a printable character, using:
```javascript
const wasPrintableChar = /[^\x00-\x1F]/.test(keyString);
```

src/scripts/choices.ts
```diff
@@ -1461,7 +1461,7 @@ class Choices implements Choices {
           the input was not focussed at the time of key press
           therefore does not have the value of the key.
         */
-        this.input.value += keyString.toLowerCase();
+        this.input.value += event.key.toLowerCase();
```
mtriff, line 1464: Was this required for the non-Latin characters to display correctly?

ousmorez: yes, Exactly.

src/scripts/choices.ts
ousmorez: Greate, Thanks.
@mtriff

mtriff: Yes, please do.

ousmorez: Done 
@mtriff

mtriff (dismissed):

ousmorez:
I checked the error in the pipeline and found that the problem is solved if the following packages are installed:
```json
  "pixelmatch": "^5.3.0",
    "puppeteer": "^19.2.2",
```
We need to add these packages to `package.json`.
This has nothing to do with my changes, should I add these packages?
@mtriff

ousmorez:
I can't resolve the failed checks, can you please help me?
@mtriff

mtriff (approved):

mtriff:
Thanks so much for the contribution @ousmorez!

PR #1154, opened 2024-05-14
Fixes an issue where passing in customProperties would render invalidly encoded object

sdglhm:
## Description

This fixes an issue where when passing in customProperties to a Choice would lead to render `[object Object]` instead of properly rendering JSON structure.

## Types of changes

## Checklist

- [x] I have added new tests for the bug I fixed/the new feature I added.

### DIFF
--- a/src/scripts/templates.ts
+++ b/src/scripts/templates.ts
@@ -321,7 +321,9 @@ const templates = {
     const opt = new Option(label, value, false, active);
 
     if (customProperties) {
-      opt.dataset.customProperties = `${JSON.stringify(customProperties)}`;
+      opt.dataset.customProperties = `${JSON.stringify(
+        customProperties || {},
+      )}`;
     }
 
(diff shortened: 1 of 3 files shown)


src/scripts/templates.ts
```diff
@@ -321,7 +321,7 @@ const templates = {
     const opt = new Option(label, value, false, active);
 
     if (customProperties) {
```
Xon, line 323: FYI, but this isn't a sufficient test for if a javascript object is empty

sdglhm: Thank you for the feedback. I'll fix the tests and update the PR.

sdglhm: @Xon I've updated the tests. Could you please let me know if this is correct. I'm not really great with JS as I majorly work on backends. Thank you.

Xon:
I've implemented a more comprehensive fix as part of #1166

PR #1161, opened 2024-06-25
Remove malicious polyfill.io usage

alanhamlett:
Polyfill.io is redirecting to malicious websites. This PR removes the optional and unnecessary polyfill.io usage.

Issue #1155, opened 2024-05-17
Custom Properties does not work when optgroup on a select


jbender0:
I'm using custom properties (inline data attribute) to display a description and thumbnail image. This works fine on a regular select field. But when I add optgroup within the select, it falls apart and only loads the option label. All custom properties are returned as undefined.

The custom property attribute is valid JSON and when I turn off the Choices script it loads correctly with optgroup just fine. So there's definitely a bug in Choices.

Here's a screenshot of it working with no optgroup and after optgroup is added (no thumbnails or custom properties).

![Screenshot 2024-05-17 at 1 38 32 PM](https://github.com/Choices-js/Choices/assets/54802325/9a664010-4de5-4678-89fa-8469d239df84)
![Screenshot 2024-05-17 at 1 40 54 PM](https://github.com/Choices-js/Choices/assets/54802325/0bafed32-02e0-4403-90e5-55999691c2e3)

Using macOS latest Chrome (v124.0). 

Anyone else ran into this bug?

PR #1152, opened 2024-03-19
#926: Maintain groups in search mode

stefan-korn:
## Description
Proposal for solving #926 
Add a new bool config option "appendGroupInSearch". If this is set to true the group of an option will be appended to the name of the option when searching.

This is somehow a workaround, because I was not able at this moment to keep the goups like in the inital state when not searching.

But at least this should help if you have the same label for options in different groups.

Remark: to run a build of the project I needed to remove the [bundlesize](https://www.npmjs.com/package/bundlesize) library. This does at least not seem to work with latest (later) node version, because it requires very old (0.1.0) version of [brotli-size](https://www.npmjs.com/package/brotli-size/v/0.1.0?activeTab=dependencies) and this requires deprecated [iltorb](https://www.npmjs.com/package/iltorb) package. Did not commit changes to package.json/package-lock.json though.

## Screenshots (if appropriate)
choices_screen
(the group City/State appended to the choice New York/Nevada/Pennsylvania)

## Types of changes

- [ ] Chore (tooling change or documentation change)
- [ ] Refactor (non-breaking change which maintains existing functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [ ] My code follows the code style of this project.
- [ ] I have added new tests for the bug I fixed/the new feature I added.
- [ ] I have modified existing tests for the bug I fixed/the new feature I added.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Issue #1132, opened 2023-09-12
Choices does not behave like other HTML elements in a disabled fieldset


tagliala:
**Describe the bug**
It is expected that HTML elements are disabled in a globally disabled `fieldset`

Ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#attributes

> If this Boolean attribute is set, all form controls that are descendants of the `<fieldset>`, are disabled, meaning they are not editable and won't be submitted along with the [`<form>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form). They won't receive any browsing events, like mouse clicks or focus-related events. By default browsers display such controls grayed out. Note that form elements inside the [`<legend>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend) element won't be disabled.

**To Reproduce**
1. Go to https://jsfiddle.net/tagliala/yetk9sgh/13/

**Expected behavior**
Choices.js appears and behave like a disabled fieldset

**Screenshots**

<img width="702" alt="image" src="https://github.com/Choices-js/Choices/assets/556268/3991883a-14f0-444b-8e9d-afa5ad4d5ff5">

**Desktop (please complete the following information):**
 - OS: macOS 13.5.2
 - Browser Chrome / Safari / Edge
 - Version: latest

**Additional context**
On Firefox, the element is not interactable but it looks like an interactable field

**Workaround**

```js
  callbackOnInit: function () {
    // Workaround for Choices-js/Choices#1132
    if (this.passedElement.element.closest('fieldset:disabled')) {
      this.disable()
    }
  },
```

Xon:
Implemented as part of #1166

PR #1166, opened 2024-08-06
v11.0.x

Xon:
### ⚠ BREAKING CHANGES
* `allowHtml` now defaults to false.
* Update to using Fuse.js v7.0.0
* Update choices.js package to be an ES module, and use '[subpath exports](https://nodejs.org/api/packages.html#subpath-exports)' to expose multiple versions (UMD, MTS bundles, can product CJS bundles but it is not enabled by default).
* Provide "fuse full" (default `choices.js`, ~19.9KB), or "fuse basic" (`choices.search-basic.js` ~18.8KB) or "prefix filter" (`choices.search-filter.js` ~14.6KB) based on how much Fuse.js is included.
* HTML escaping of choice/item labels should no longer double escape depending on allowHTML mode.
* Templates/text functions now escape `'` characters for display.
* `addItemText`/`uniqueItemText`/`customAddItemText` are now called with the `value` argument already escaped.
* Typescript classes for input data vs internal working data have been adjusted resulting in the `Choice`/`Group`/`Item` typescript classes have been renamed, and aliases left as required.
* For `select-one` and `select-multiple`, the placeholder value is pulled from `config.placeholderValue="..."` or `<select data-placeholder="...">` before attempting to extract a placeholder from the options list. [#912](https://github.com/Choices-js/Choices/issues/912) [#567](https://github.com/Choices-js/Choices/issues/567) [#843](https://github.com/Choices-js/Choices/issues/843)
* Typescript source mapping files `.d.ts.map` are no longer generated and various `.test.d.ts` files are no longer generated
* Improve consistency of the `choice` event firing. `choice` event now occurs after the `addItem` event
* `enter` key now consistently opens/closes the dropdown instead of the behavior varying depending on backing element or internal state of the highlighted choice
* Mutation APIs `setChoiceByValue`/`setChoices`/`setValue` now throw an error if the Choices instance was not initialized or multiple choices instances where initialized on the same element. Prevents bad internal states from triggering unexpected errors [#1129](https://github.com/Choices-js/Choices/issues/1129)
* Trigger a search event (with empty value and 0 resultCount) when search stops

### Features

* Improve performance of search/filtering with large number of choices.
* Add `closeDropdownOnSelect` option, controls how the dropdown is close after selection is made. [#636](https://github.com/Choices-js/Choices/issues/636) [#973](https://github.com/Choices-js/Choices/issues/873) [#1012](https://github.com/Choices-js/Choices/issues/1012)
* Allow choices.js to be imported on nodejs, useful for tests and also server side rendering. As windows.document is by default not defined, the default template rendering will not function. The `callbackOnCreateTemplates` callback must be used. [#861](https://github.com/Choices-js/Choices/issues/861)
* `config.classNames` now accept arrays to support multiple classes. [#1121](https://github.com/Choices-js/Choices/issues/1121) [#1074](https://github.com/Choices-js/Choices/issues/1074) [#907](https://github.com/Choices-js/Choices/issues/907) [#832](https://github.com/Choices-js/Choices/issues/832)
* The original option list for the select is not destroyed, and all loaded choices are serialised to HTML for better compatibility with external javascript. [#1053](https://github.com/Choices-js/Choices/issues/1053) [#1023](https://github.com/Choices-js/Choices/issues/1023)
* New `singleModeForMultiSelect` feature to treat a `select-single` as if it was a `select-multiple` with a max item count of `1`, and still auto-close the dropdown and swap the active item on selection.
* `Remove item text` can be localized.
* Allow user-created choices for selects. [#1117](https://github.com/Choices-js/Choices/issues/1117) [#1114](https://github.com/Choices-js/Choices/issues/1114)
    * User input is escaped by default. At the risk of XSS attacks this can be disabled by `allowHtmlUserInput`.
* Render options without a group even if groups are present. [#615](https://github.com/Choices-js/Choices/issues/615) [#1110](https://github.com/Choices-js/Choices/issues/1110)
* Read `data-labelclass`/`data-label-description` from `<option>` HTML to drive adding a per-choice CSS label and description text when `allowHtml: false`.
* Add `removeItemButtonAlignLeft` option, to control if the remove item button is at the start or the end of the item.
* Add `removeChoice` method. Removes the choice from the `choices.js` object and any backing `<option>` HTML element
* Add `refresh` method. Reloads choices from the backing `<select>`s options.
* `escapeForTemplate` function is passed to the 2nd method of the `callbackOnCreateTemplates` callback.
* When `allowHtml` is false, default templates now render escaped html to `innerHtml` writing to `innerText`.
    * This provides consistent rendering performance as `innerText` is quirky and slower than escaped html into `innerHtml`
* Shadow DOM support [#938](https://github.com/Choices-js/Choices/pull/938)
* `searchResultLimit` can be set to `-1` for no limit of search results to display.

### Bug Fixes
* Avoid pushing a search to `fuse.js` which is just additional whitespace to the existing search term
* Replace malicious polyfill with cdnjs. [#1161](https://github.com/Choices-js/Choices/issues/1161)
* Maintain groups in search mode. [#1152](https://github.com/Choices-js/Choices/issues/1152)
* Fix various "first press" bugs on single select dropdowns. [#1104](https://github.com/Choices-js/Choices/issues/1104)
* Fix 'esc' would close the dropdown and also apply to the container resulting in an overlay/modal unexpectedly closing. [#1039](https://github.com/Choices-js/Choices/issues/1039)
* Fix form reset would clear the choices list, but not clear the search bar. [#1023](https://github.com/Choices-js/Choices/issues/1023)
* Fix options would be disabled when choices.js was initialized on a disabled `<select>` element. [#1025](https://github.com/Choices-js/Choices/issues/1025)
* Fix a `search_term` element to appear in form submit data. [#1049](https://github.com/Choices-js/Choices/issues/1049)
* Fix 'remove item' button would trigger the change event twice due to placeholder value being used (match html single-select). [#892](https://github.com/Choices-js/Choices/issues/892)
* Fix optgroups are not preserved when Choices is destroyed [#1055](https://github.com/Choices-js/Choices/issues/1055)
* Fix placeholder config option would be ignored for select boxes which have blank entries.
* Fix `data-custom-properties` attribute did not serialize to created elements as a json blob as expected. [#840](https://github.com/Choices-js/Choices/issues/840) [#1155](https://github.com/Choices-js/Choices/issues/1155) [#543](https://github.com/Choices-js/Choices/issues/543)
* Fix multi-select did not correctly resizing when a select option is selected on choices.js initialization.
* Fix `clearInput` function did not clear the last search.
* Fix `addItemFilter` would allow empty strings as input to be added for items.
* Fix various issues with double escaping when displaying items/choices depending on allowHTML mode.
* Fix `aria-label` for placeholders was set to the string `null`
* Fix `searchEnable` flag was not respected for `select-multiple` [#1042](https://github.com/Choices-js/Choices/issues/1042)
* Fix poor error message when Choices is passed a string selector which fails to find the element for Choices to attach to.
* Improve various `[aria-*]` attribute handling for better lighthouse accessibility scores [#1169](https://github.com/Choices-js/Choices/issues/1169)
* Improve contrast on default CSS by darkening primary item selection color [#924](https://github.com/Choices-js/Choices/issues/924)
* Fix Choices does not accept an element from an iframe [#1057](https://github.com/Choices-js/Choices/issues/1057)
* Fix Choices was not disable in a `<fieldset disabled>` [#1132](https://github.com/Choices-js/Choices/issues/1132)
* Fix `silent` option does not silence warnings about unknown options [#1119](https://github.com/Choices-js/Choices/issues/1119)
* Fix documentation that suggests duplicateItemsAllowed works with select-multiple, when it only works for text. [#1123](https://github.com/Choices-js/Choices/issues/1123)
* Fix quadratic algorithm complexity (aka O(N^2) ) when filtering/search choices.
* Fix search results could be unexpectedly unstable, and that `fuseOptions.sortFn` was effectively ignored [#1106](https://github.com/Choices-js/Choices/issues/1106)
* Fix `select-one` placeholder could ignore the non-option placeholder configuration
* Fix search did not trigger to copy&paste events [#860](https://github.com/Choices-js/Choices/issues/860) [#174](https://github.com/Choices-js/Choices/issues/174)
Fix the page scrolls when you press 'space' on a single select input [#1103](https://github.com/Choices-js/Choices/issues/1103)

### Chore
* Remove various unused code
* For test html pages, prevent a failing `fetch()` from breaking the rest of the examples
* Tweak `_render` loop to avoid duplicating has-changed checks
* Switch from `webpack` to `rollup`,  and provide `esm`/`cjs`/`umd` bundles by default. Enables tree shaking! ~3kb dropped from the `umd` bundle (aka compatible with the previous produced webpack bundle)
* Switch from `mocha` to `vitest` as the test running framework.
* Switch from `puppeteer`/`selenium`/`cypress` to `playwright` as the end-to-end test running framework.
* Restructure end-to-end tests so html/script blocks are co-located to improve debugability
* Enable `@typescript-eslint/explicit-function-return-type` eslint rule
* Remove `deepMerge` dependency.

Tests:
- [x] TS/JS linting
- [x] css linting
- ~~[x] selenium (macos-latest, firefox)~~
- ~~[x] selenium (windows-latest, edge)~~
- ~~[x] selenium (windows-latest, firefox)~~
- ~~[x] selenium (windows-latest, chrome)~~
- ~~[x] puppeteer~~
- [x] Address various unit tests failures

playwright end-to-end tests:
- [X] text
- [X] select-one
- [x] select-multiple
- [x] replace puppeteer test with playwright
- [X] replace selenium test with playwright
- [X] debug playwright github action

playwright e2e test run against `choices.js` and `choices.min.js` bundles. Extending to support different bundles is easy.

playwright supports the following browsers:
- chromium
- firefox
- webkit

playwright supports the following OS's:
- mac os
- linux
- windows


Xon:
Broken test configuration:
- [ ] bundle size - expired credentials, need replacing
- [ ] end-to-end tests. Typescript integration is failing
- [ ] unit tests. Typescript integration is failing

Failing tests:
- [ ] Various browser tests, differences in screenshots need to workout how to generate them and then compare.

Xon:
Updated to using Fuse.js v7.0.0, and we can now export bundles based on fuse full or fuse basic, with a "search-prefix" as a baseline reference;

```
choices.min.js: 20.36KB
choices.search-basic.min.js: 19.31KB
choices.search-prefix.min.js: 15.27KB
```

Xon:
Plan is to replace the selenium/puppeteer/cypress e2e test with (Playwright)[https://github.com/microsoft/playwright]. 

This should provide a single e22 test framework and hopefully improve reliability of some of these tests

Xon:
Test run against choices.js and choices.min.js. Extending to support different bundles is easy.

playwright supports the following browsers:
- chromium/firefox/webkit
playwright supports the following OS's:
- mac os/linux/windows

New end-to-end tests:
- [X] text
- [X] select-one
- [ ] select-multiple
- [ ] replace puppeteer test with playwright
- [ ] replace selenium test with playwright

Xon:
This matches the previous e2e tests and unit tests. 

I plan to add a couple more e2e tests to cover features added, and need to update the change notes attached to the PR to match the latest changelog, but I think this is getting closer.

PR #1170, opened 2024-08-22
Draft: Large item count performance

Xon:
Performance for large number of items and choices is lacking.

The following bottlenecks have been identified:
- No end-to-end tests to identify performance issues with large number of items
   - [X] Example e2e page experiencing performance issues
   - [X] e2e test against the page to track and confirm the issues are resolved
- [X] All items are re-rendered whenever an item is added or removed
   - track the rendered element, and only render it if required


Xon:
I've gotten item and choices rendering performance being acceptable, but selecting a choice and it being converted into an item is still doing more work than expected

PR #1188, opened 2024-08-31
Fix regression where webpack doesn't permit importing scss/css

Xon:


Xon:
Abandoned in favor of #1193

PR #1192, opened 2024-08-31
Fix regression "no choices to choose from"/"no results found" notice did not reliably trigger

Xon:


tagliala:
Hi,

I'm testing this branch

It works better than the actual implementation, but there are still a few issues with "dynamic" fields.

I think it is related to some missing checks after `setChoices` 

v11: https://jsfiddle.net/tagliala/rx8mtbz6/17/

image

v10: https://jsfiddle.net/tagliala/rx8mtbz6/18/

image

Xon:
@tagliala It looks like `setChoices` didn't clear the cached notice, so the last notice "stuck". That should be fixed.

I think the entire notice system needs a bit of a rework to avoid this ad-hoc matching and to be more centralized (basically making a notice reducer which listens for the relevant add/remove events.

But for now the current code should work

tagliala:
@Xon thanks, can you rebase this branch on master? I would like to take a look

PR #1240, opened 2024-12-21
Fix search was not stopped when leaving focus with `esc` key

zoglo:
## Description

This PR aims to refresh the list when clicking escape if a search is active.
There currently is a bug where you'd search for something and want to focus out of it when pressing escape - when this happens, you'll lose focus completely and there will not be any items left with a blank search input.

This can be reproduced in the current demo when searching for something and pressing escape - the choices will be gone and only appear again if you type something and then delete it again:
https://choices-js.github.io/Choices/

-> See Screenshot below

image

## Types of changes

## Checklist

### DIFF
--- a/src/scripts/choices.ts
+++ b/src/scripts/choices.ts
@@ -1762,6 +1762,7 @@ class Choices {
     if (hasActiveDropdown) {
       event.stopPropagation();
       this.hideDropdown(true);
+      this.refresh(false, false, true);
       this.containerOuter.element.focus();
     }
   }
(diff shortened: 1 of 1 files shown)


Xon:
`this.refresh(false, false, true);` is probably the wrong function to call here, as it it rebuilds the internal state from the raw `<select>`'s options list which is much more heavy weight than required.

I think swapping for `this._stopSearch();`

zoglo:
> `this.refresh(false, false, true);` is probably the wrong function to call here, as it it rebuilds the internal state from the raw `<select>`'s options list which is much more heavy weight than required.
> 
> I think swapping for `this._stopSearch();` will make this work as expected

Yeah, just tested it and it works as espected - good catch, didn't see that method :')

PR #1242, opened 2024-12-22
Fix `tab` => direction keys handling in single-select mode

Xon:


public/assets/scripts/choices.js
```diff
@@ -4880,17 +4887,11 @@
...
-                else {
+                if (targetIsInput) {
                     containerOuter.removeFocusState();
-                    if (targetIsInput || (target === containerOuter.element && !this._canSearch)) {
-                        this.hideDropdown(true);
+                    this.hideDropdown(true);
+                    if (this._isTextElement || this._isSelectMultipleElement) {
+                        this.unhighlightAll();
                     }
                 }
```
zoglo, line 4896: Whilst this works for the search not being hidden, we now don't take into account to remove the focusState on the previous outerElement when tabbing through. The previous element won't lose their `is-focused` class.

We'd actually need to check for another condition afterwards to remove the previous class:


```js
if (target === this.input.elemen) {
   // your logic
}
// Remove the focus state when the past outerContainer was the target
else if (target === this.containerOuter.element) {
  containerOuter.removeFocusState();
}
```

zoglo (changes requested):
This one was on my list as well when the search couldn't be used. There's actually one more issue that is now introduced which is the focus class not being removed onBlur - I left a possible solution / case in my comment

public/assets/scripts/choices.js
Xon: I've restored that branch, and compiled the js file. Want to give that a test to see if that encounters any other regressions or accessibility issues?

zoglo: Already did - the change looks good to me :)

PR #1258, opened 2025-01-29
Added clear event

phaedra-casadio-dotbeyond:
Added in EventType constants the "clear" event.


Xon:
@phaedra-casadio-dotbeyond can you add this PR as part of #1259 ? This ensures the automated changelog notes aren't full of unrequired chatter

PR #1270, opened 2025-02-21
Fix: Reached maximum item limit notice is not cleared after removing selections

### LINKED ISSUE #1249: In multiselect "Only 3 values can be added" is displayed even if there are choices
**Describe the bug**
If you select all possible choices (and you add `max-item-count="3"`), dropdown will show "Only 3 values can be added", which is correct. But if you remove at least one value, text is still visible, even if there value to choose from.
image

**To Reproduce**
1. Create Multiple select input with some values
2. Add `max-item-count="3"` params
3. Select 3 choices
4. Remove 1 or more choices
5. Open select again and there is still text "Only 3 values can be added", see screenshot above.

**Expected behavior**
Text "Only 3 values can be added" should not be displayed when there are values, that can be selected.

**Choices version and bundle**
- Version: 11.0.3

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Chrome

Thanks in advance for your help

ticktackk:
closes #1249


Xon (approved):

Xon (changes requested):
Various E2E tests are failing which need addressing

Xon (approved):

Issue #1281, opened 2025-02-26
11.0.4/11.0.5: Regression with `addItem` + `clearChoices`?


tagliala:
**Describe the bug**
The change introduced in Choices-js/Choices@ac58874 as part of Choices-js/Choices#1272, introduced a regression with a combination of `addItem + clearChoices`

### v11.0.3
This was working in 11.0.3: https://jsfiddle.net/tagliala/0mzjok8g/7/

![Image](https://github.com/user-attachments/assets/d86f69f4-e31a-4ad1-b969-3c747c28f705)

### v10.2.0
According to #1272, the intent was to fix a regression from 10.x, but as per this approach had the same behavior in 10.x: https://jsfiddle.net/tagliala/0mzjok8g/10/

![Image](https://github.com/user-attachments/assets/3b0316c7-eb54-4382-8a27-6c32bf0006a5)

**To Reproduce**
1. Go to https://jsfiddle.net/tagliala/0mzjok8g/5/
2. Click on the select box
3. Select "Test"

**Expected behavior**
"Test" being selected, as in 11.0.3

**Screenshots**

![Image](https://github.com/user-attachments/assets/f7780c00-06dc-460f-8a03-5e5dc1f705f9)

**Choices version and bundle**
- Version: v11.0.4 and v11.0.5 choices.min.js

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: all tested
 - Version: latest

**Smartphone (please complete the following information):**
didn't test

**Additional context**
I'm not sure that this is a regression. In fact, I expect that `clearChoices` actually clears also the selected choice, so this should be the correct behavior

Opening an issue the same, but please feel free to close if this is the expected behavior according to you too

Xon:
PR #1283 should fix this, and importantly adds some e2e tests to capture this behavior. Only trick is there is a slight behavior change for `select-single` vs `select-multiple` in the PR compared to `v10`.

Xon:
`v11.0.6` should fix the reported issue

Issue #1278, opened 2025-02-24
11.0.4 - setChoices when replaceChoices is true, the default selected is cleared


jackbayliss:
**Describe the bug**
When using setChoices, as of 11.0.4 if you set replaceChoices to true, the default selected option is removed and no longer visible - where as on 11.0.3 it showed the default selected, even if it was true. 

**To Reproduce**
1. Visit https://jsfiddle.net/mn03b8jh/2/ (basic implementation)
2. Witness the select is not preselected with the selected value
3. Change the js script from 11.0.4 to 11.0.3 in the fiddle and click run
4. It works with 11.0.3

**Expected behavior**
I would expect it to be the same between 11.0.3 and 11.0.4 as it wasn't a major release --unless it was? 
This may be the new expected behaviour, but wanted to raise just in case, as it caught me by surprise.

tagliala:
Hello,

we are also experiencing a major regression in 11.0.4 compared to 11.0.3, but I'm not sure that it is the same issue.

It is also more difficult to replicate, I'll follow this issue and see if the fix here will fix our use case too

Xon:
I believe this would have been related to a performance fix to avoid accidently quadratic behavior when clearing choices https://github.com/Choices-js/Choices/commit/f9a8499e9af55f9ae66459697014eb4a4ad827c5. If you need the placeholder to persist between setting choices `data-placeholder` attribute should be the most easily usable.

I'm open to suggestions on the best way to resolve this.

tagliala:
Thanks for the quick fix.

Unfortunately, as I suspected, the regression we've experiencing is not fixed by 11.0.5

At the moment I do not have time to provide a reproducible test case. We'll stay on 11.0.3 and provide further details if we can allocate some time to debug

Xon:
`setChoices()` with `replaceChoices: true` has had fairly inconsistent behavior for `v11` :(
  - `select-single` mode:
    - In `v10.2.0`/`v11.0.0` - selection is preserved, and selection+new choices does not result in duplicates in the choices list
    - In `v11.0.1` - the selection is not preserved, and selection+new choices does not result in duplicates in the choices list
    - In `v11.0.2`/`v11.0.3` - the selection is preserved, and selection+new choices can results in duplicates in the choices list
    - In `v11.0.4`/`v11.0.5` - the selection is not preserved, and selection+new choices can results in duplicates in the choices list
  - `select-multiple` mode:
[...]

PR #1283, opened 2025-02-26
Fix setChoices & clearChoices related regressions

Xon:
- Fix 11.0.4/11.0.5: Regression with addItem + clearChoices #1281
- 2nd attempt at fixing #1278 to remove regressions. See https://github.com/Choices-js/Choices/issues/1278#issuecomment-2686343993 for details on how 
- Adds e2e tests to cover both cases


tagliala:
Confirming this fixes #1281 without changes to the existing code

Issue #941, opened 2021-04-06
Aria issues


viotopo:
I am having to issues validating the code and I don't know how to do it with choices.

1. ARIA items do not have accessible names

2. Elements with an ARIA [role] that require children to contain a specific [role] are missing some or all of those required children

Anybody knows how I can implement this on choices?

Greg-Boggs:
We are also seeing critical accessibility bugs with Aria tags and chosen. Interestingly, these are only see in the Drupal implementation of Chosen and not in the Chosen Demo.

To replicate this, you have to install the drupal module, and configure it to display in Drupal 9, and then scan the page with an accessibility scanner like Axe.

Here's the correct place to follow up with the Drupal bug

https://www.drupal.org/project/chosen/issues/2834096

Greg-Boggs:
Forgive me. I have replicated the errors using Axe scanner on the demo.  If you scan the demo you'll see the WCAG 2.0 AA aria violations.

https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd

seredniy:
Hey all!

Any news here?

Xon:
The v11.0.x updates did see some aria related changes but it isn't completely fixed. Should be better than it used to be.

mlinnetz:
Additionally, if we're entertaining minor a11y changes, a typical Choices dropdown will add the attribute 
`role="textbox"`
to the search textbox:
`<input type="search" class="choices__input" role="textbox" >`

According to a11y best practice, it's recommended to **not** add the role="textbox" to an <input> text or search element . Rather, screen readers and other assistive technologies already infer that it's a textbox from the <input> element itself.

P.S. Thanks to the collaborators.  Choices.js is a great plugin.

Xon:
That should be a simple PR (looks to be just one spot adding the html attribute, and then updating a unit-test). Failing that can you link to the relevant a11y best practice so the change could be documented?

mlinnetz:
> That should be a simple PR (looks to be just one spot adding the html attribute, and then updating a unit-test). Failing that can you link to the relevant a11y best practice so the change could be documented?

I believe this is the precise link on the w3C aria documentation:
https://www.w3.org/TR/wai-aria-1.2/#textbox
and the relevant passage is:

[...]

PR #1285, opened 2025-03-03
Remove `role="textbox"` from search input, per a11y practices. #941

mlinnetz:
## Description
A typical Choices dropdown will add the attribute role="textbox" to the search textbox:
`<input type="search" class="choices__input" role="textbox" >`

According to a11y best practice, it's recommended to not add the role="textbox" to an text or search element . Rather, screen readers and other assistive technologies already infer that it's a textbox from the element itself.

https://www.w3.org/TR/wai-aria-1.2/#textbox
> The intended use is for languages that do not have a text input [element](https://www.w3.org/TR/wai-aria-1.2/#dfn-element), or cases in which an element with different [semantics](https://www.w3.org/TR/wai-aria-1.2/#dfn-semantics) is repurposed as a text field.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/textbox_role
> The textbox role is used to identify an element that allows the input of free-form text. Whenever possible, rather than using this role, use an [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) element with [type="text"](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text), for single-line input, or a [<textarea>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) element for multi-line input.

In other words, both sources imply this should only be used on non-input elements.  

## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)

This PR removes setting a role="textbox" to the **input type="search"** element of a choices dropdown.

## Checklist

- [X] I have modified existing tests for the bug I fixed/the new feature I added.

### DIFF
--- a/public/assets/scripts/choices.js
+++ b/public/assets/scripts/choices.js
@@ -3247,7 +3247,6 @@
             inp.autocomplete = 'off';
             inp.autocapitalize = 'off';
             inp.spellcheck = false;
-            inp.setAttribute('role', 'textbox');
             inp.setAttribute('aria-autocomplete', 'list');
             if (placeholderValue) {
                 inp.setAttribute('aria-label', placeholderValue);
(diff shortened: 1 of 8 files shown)


Xon (approved):
LGTM

PR #1260, opened 2025-02-02
Fix `tab` => direction keys handling with disabled search

zoglo:
## Description

This is a continuation of #1242

Whilst the PR in #1242 did work, we forgot the case that a choices may not have an input field when the search is disabled.
Without a search input, we are not closing the dropdown if it has been open when tabbing away.

## Types of changes

- [ ] Chore (tooling change or documentation change)
- [ ] Refactor (non-breaking change which maintains existing functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [x] My code follows the code style of this project.
- [ ] I have added new tests for the bug I fixed/the new feature I added.
- [ ] I have modified existing tests for the bug I fixed/the new feature I added.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1308, opened 2025-05-30
Use `config.searchEnabled` instead of `_canSearch` in the `_onBlur` method

leofeyer:
## Description

This is a follow-up on #1260.

We have to use `this.config.searchEnabled` instead of `this._canSearch` because the latter is sometimes set to `false`, e.g. in the `_onSelectKey` and `_onDirectionKey` methods.

https://github.com/Choices-js/Choices/blob/8c0c11e26f2d1446263c20f85e0c437c6e37f74e/src/scripts/choices.ts#L1806

The fix should probably be verified by @zoglo before you consider merging it.

## Types of changes

## Checklist


zoglo:
Reproduction of this is:

1. Open any Select Singlebox with a search
2. Type in something in the search, then delete it
3. Choose an option with an arrow key
4. With your mouse, focus out to any other field (be it a select or input)
5. Use the keyboard to navigate back and press "down" to get to the first option

We now have a state where choices instantly closes and when reopening, the search can not be focused anymore:

https://github.com/user-attachments/assets/55f3d404-783d-44e0-8a17-9596c97c50c8

PR #1315, opened 2025-06-29
Disable `searchResultLimit` by default

fritzmg:
## Description

I got confused when using some of the default settings of Choices.js plus `searchEnabled: true` that I could not find entries via the search in larger selects, unless I typed in more letters of the entry that I was looking for. So for example if I typed the letter _a_, it did not show all entries with that letter.

The reason for this is that the default `searchResultLimit` is set to _4_:

https://github.com/Choices-js/Choices/blob/8c0c11e26f2d1446263c20f85e0c437c6e37f74e/src/scripts/defaults.ts#L60

However, this does not make much sense to me as by default the `renderChoiceLimit` is set to `-1`

https://github.com/Choices-js/Choices/blob/8c0c11e26f2d1446263c20f85e0c437c6e37f74e/src/scripts/defaults.ts#L41

i.e. Choices.js will always show all entries of the `<select>`. But when searching for something - which would already filter down the entries to less or equal than _all_ - the result is suddenly limited to 4.

So I think it makes sense that the default for `searchResultLimit` should always be the same as `renderChoiceLimit`.

Generally I am not sure what the use-case is for having a `searchResultLimit` that is lower than the `renderChoiceLimit`.

## Types of changes

I think this constitutes a breaking change because if someone configured a `renderChoiceLimit` because they have a lot of `<option>`s and they did not set a `searchResultLimit`, they would suddenly potentially get more search results than their `renderChoiceLimit`.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ import "choices.js/public/assets/styles/choices.css";
     searchEnabled: true,
     searchChoices: true,
     searchFloor: 1,
-    searchResultLimit: 4,
+    searchResultLimit: -1,
     searchFields: ['label', 'value'],
     position: 'auto',
     resetScrollPosition: true,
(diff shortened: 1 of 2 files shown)


fritzmg:
But the list can never be larger than `renderChoiceLimit`, can it?

PR #1330, opened 2025-07-26
Define [aria-selected] for selectable choices per WAI-ARIA 1.2 spec, and avoid triple state with aria-selected

pablobm:
## Description

The other day I had some strange styling going on with a CSS rule targeting `[aria-selected]` on a multiselect. The issue was that `aria-selected` is initially not present on any option, then on highlight it's set to `aria-selected="true"`, then when out of highlight it changes to `aria-selected="false"`. Therefore my CSS rule should have targeted `[aria-selected=true]` instead (or perhaps `.is-highlighted`).

Although I fixed my issue, this left me thinking if this tri-state should be desirable at all. Wouldn't it make more sense to either start from `aria-selected="false"` or to unset `aria-selected` when not highlighted?

Looking at the [relevant section of the WAI-ARIA 1.2 spec](https://www.w3.org/TR/wai-aria-1.2/#aria-selected) I found the following table:

| Value               | Description                             |
| ------------------- | --------------------------------------- |
| false               | The selectable element is not selected. |
| true                | The selectable element is selected.     |
| undefined (default) | The element is not selectable.          |

I'm not an ARIA expert and I know this stuff is difficult to get right. With that said, I want to raise the question with this PR: should options be `aria-selected=false` by default?

## Types of changes

## Checklist

- [x] I have added new tests for the bug I fixed/the new feature I added.

I wasn't sure how to go about testing this. After looking at similar ones, I decided to create a dedicated `describe` block, which may grow with other tests in the future. Thoughts?


Xon:
I agree setting `aria-selected="false"` on choices which aren't selected would make sense.

However, where you are doing `div.setAttribute('aria-selected', 'false');` isn't correct.

Xon (dismissed):

Issue #1131, opened 2023-09-11
Behaviour of dropout elements on mobile device


Slisarenko:
The behavior on a mobile device isn't very clear, I have an iPhone 14 Pro.
The situation is that the items from the input fall under the bottom of the container, which are at the bottom.
For laptop - items are displayed on top of the container below.
Maybe someone has ideas, what could be the problem and what should be paid attention to?
![IMG_1027](https://github.com/Choices-js/Choices/assets/55946144/4105ad01-3822-4400-a709-207552e8b1b2)

https://github.com/Choices-js/Choices/assets/55946144/3092ce23-70db-49fd-a7f7-50daab480c2d

Xon:
The dropdown being obscured by the container is a known issue, and I'm open to PR's to help improve the CSS which is causing this.

PR #1342, opened 2025-08-23
Implement a `dropdownParent` feature

### LINKED ISSUE #930: dropdown visibility height
Hi,

First, many thanks for this component !

I'm not sure it's a bug but when dropdown is hidden, it still makes the height of the parent as tall as the dropdown even if hidden.
The only solution I found was to set dropdown height to 0 and when .is-active, put the height to auto.
Is this the way to go or maybe there is another more elegant way please ?

Thanks for your attention,
Regards

### LINKED ISSUE #1288: Dropdown parent must be configurable to avoid overflows
When Choices is put into a container with overflow hidden, drowdown is truncated. This because the dropdown element is inside the component.

I need a dropdownParent configuration.

### LINKED ISSUE #1309: Dropdown - Body appendTo option
Hi Team,

I kindly request you to consider this feature addition, as it addresses a major usability issue.

Currently, when we use **Choices.js** inside a responsive table (scrollable table), the dropdown is not fully visible outside the table's scroll area. This happens because the dropdown is appended near the input field within the scrollable container, causing it to be clipped.

To resolve this, we need an option that allows the dropdown to be appended to the <body> instead of the input field container. Additionally, the dropdown’s position should be automatically calculated and adjusted based on the input’s location in the viewport.

This change would greatly improve usability in responsive layouts and prevent overflow issues.

Thank you for your consideration.

Best regards,
Unnikrishnan PJ

### LINKED ISSUE #1340: What's the equivalent to Select2's DropdownParent: 'body in Choices?
I have some elements that are positioned to have a fixed height, etc and trying to use Choices JS the dropdown gets cut off. The solution for that in Select2 was to use DropdownParent: 'body" to make sure such dropdowns are added to body and float nicely above everything else.

How would I achieve something like this with Choices?

### LINKED ISSUE #1375: appendTo option for dropdown (Must Feature) (e.g. before closing body / near to the dropdown)
Currently, the Choices dropdown is always rendered within the **.choices wrapper**

When the field is placed inside a scrollable container, modal, or element with **overflow: auto/hidden**, the dropdown gets clipped and becomes partially unusable. Please find the below screenshot

Image

Many UI libraries (Select2, Popper-based components, etc.) provide an **appendTo** / portal option to render the dropdown outside the local DOM context (commonly document.body) while positioning it relative to the input.

Adding a similar option in Choices would greatly improve usability in complex layouts.

--------------------------------------------------------------------------

**Why this is a must-have**

In modern UIs, forms are commonly placed inside:

- Modals
- Side panels
- Scrollable dashboards
- Fixed / sticky layouts

Without an **appendTo** option, developers are forced to implement fragile workarounds (manually moving the dropdown to body, recalculating position, handling scroll/resize), which defeats the purpose of using a library.

Most comparable select libraries provide this capability out of the box.

zoglo:
## Description

The implementation would allow specifying a `dropdownParent` and if provided, all dropdowns would be appended to it. Additionally, each dropdown is still positioned absolute to the parent (I scratched the idea of `position: fixed` in favor of following the KISS principle and positioning calculation).

This would implement and fix the following:

- https://github.com/Choices-js/Choices/issues/930
- https://github.com/Choices-js/Choices/issues/1131
- https://github.com/Choices-js/Choices/issues/1288
- https://github.com/Choices-js/Choices/issues/1309
- https://github.com/Choices-js/Choices/issues/1340

This isn't a feature I need right now (not in the next release and I don't really want to delay that one 🙈), but I experimented with implementing it. It still requires significant selector adjustments for the CSS to work correctly (currently, dropdown styles aren’t fully applied). 

Maybe someone else might want to build on this functionality in the future. This PR could serve as a head start and provide some documentation for that purpose.

image

## Types of changes

## Checklist


Xon:
@zoglo I've merged in the most recent `main` and fixed the unit tests. I can definitely see that the `showDropdown()` changes look rather raw

zoglo:
> @zoglo I've merged in the most recent `main` and fixed the unit tests. I can definitely see that the `showDropdown()` changes look rather raw

It does, I know. Afair, it needed more work on the positioning, the CSS + keyboard navigation.
It definitely isn't a task that is done in a few minutes.

But thanks for checking out @Xon

tnelson0406:
eagerly awaiting this feature :)

zoglo:
(Just a reminder - And I saw the last message -  I will continue working on this when I find some time, that however won't be within the next weeks 😞)

Xon:
@zoglo when you resume work on this, you'll probably want to rebase onto latest `main` branch as I've updated vitest to latest and fixed some unit test bugs

PR #1339, opened 2025-08-22
Highlight the first selected choice within the dropdown

zoglo:
## Description

This one was on my list for a while but I had to check where to apply this one as easy as possible.
Basically this highlights the first selected choice in the dropdown on init / rendering of the items

This will fix all of these (as they are duplicates)

* https://github.com/Choices-js/Choices/issues/586
* https://github.com/Choices-js/Choices/issues/850
* https://github.com/Choices-js/Choices/issues/1290

## Types of changes

- [ ] Chore (tooling change or documentation change)
- [ ] Refactor (non-breaking change which maintains existing functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [x] My code follows the code style of this project.
- [ ] I have added new tests for the bug I fixed/the new feature I added.
- [ ] I have modified existing tests for the bug I fixed/the new feature I added.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1361, opened 2025-11-27
Improve UX on the select dropdown

zoglo:
#1339 introduced the selected option being correctly highlighted when opening the dropdown but did not account for the option also being visible when opening the dropdown, leading to bad UX (unsure if selected or not).

This PR thus introduces the following bug fixes and some other behavior changes to improve UX and make the select act more native:

1. When opening the select dropdown, we now have a native scroll to the first selected option.

    -  I am using the native `scrollToElement` function here to avoid any smooth scroll or animated scroll.

2. The selected state will always be visible; the hover (highlighted) state will be additionally visible.

    - Previously when hovering over any item, we lost focus of the checked options. This has now been changed by making use of the _yet unused_ `is-selected` CSS class. It should now match the same behaviour as other libraries (e.g., SlimSelect). I also made sure to exclude the "Press to select" from the styles on already selected options, as this hint does not make sense here.

3. When closing the dropdown, the hovered (highlighted) choices will be reset, matching the native `<select>` behaviour.

    - We previously left the last hovered choices highlighted, which could confuse users when reopening the dropdown not showing the selected option.

The following video should show all behavioral changes

https://github.com/user-attachments/assets/d633a7ab-a723-40b1-8a12-589695f3253f

## Types of changes

## Checklist

### DIFF
--- a/public/types/src/scripts/choices.d.ts
+++ b/public/types/src/scripts/choices.d.ts
@@ -200,10 +200,6 @@ declare class Choices {
         target: HTMLInputElement | HTMLSelectElement;
     }): void;
     _onInvalid(): void;
-    /**
-     * Removes any highlighted choice options
-     */
-    _removeHighlightedChoices(): void;
     _highlightChoice(el?: HTMLElement | null): void;
     _addItem(item: ChoiceFull, withEvents?: boolean, userTriggered?: boolean): void;
     _removeItem(item: ChoiceFull): void;


test-e2e/__screenshots__/chromium-win32.png
zoglo: Unrelated to this PR but the icons for the windows runner did change, just a chore to make all tests pass:
image

PR #1366, opened 2025-12-31
Fix `searchEnabled` being disabled for `select-multiple` did not work

Xon:


Xon:
e2e tests are being flaky :(

PR #1374, opened 2026-01-31
Fix input collapses for multi select & no placeholder & no initial selection

### LINKED ISSUE #1368: Input collapses for multi select & no placeholder & no initial selection
When using multi select without a placerholder or an initial value the input box seems to collapse. As far as I can tell this is caused by `setWidth()` being called which results in `1ch` as the width.
It seems like this got introduced here (https://github.com/Choices-js/Choices/commit/9fc283c5f27028b19e844b387abea5671432d05e) in order to handle initial selections.

https://github.com/Choices-js/Choices/blob/5e25576e07d9bf8310a5245773bf11bf3a7d12bc/src/scripts/choices.ts#L2296-L2309

(present in v11.1)

muendlein:
## Description

Fixes https://github.com/Choices-js/Choices/issues/1368

## Types of changes

## Checklist

### DIFF
--- a/src/scripts/choices.ts
+++ b/src/scripts/choices.ts
@@ -2305,7 +2305,10 @@ class Choices {
       if (this._placeholderValue) {
         this.input.placeholder = this._placeholderValue;
       }
-      this.input.setWidth();
+
+      if (this._placeholderValue || this._store.items.length > 0) {
+        this.input.setWidth();
+      }
     }
 
     this._highlightPosition = 0;


src/scripts/choices.ts
```diff
@@ -2305,7 +2305,10 @@ class Choices {
       if (this._placeholderValue) {
         this.input.placeholder = this._placeholderValue;
       }
-      this.input.setWidth();
+
+      if (this._placeholderValue || this._store.items.length > 0) {
```
Xon: Any reason to not simplify this check to just?
`else if (this._store.items.length) {`

muendlein: I'm not sure if I fully understand your suggestion but `if (this._store.items.length)` should always evaluate to true as `items` are initialized as `[]`.

Xon:
Looking into this more, removing the logic which reduces the input to 1ch causes reflow issues when multiple items are selected on a narrow enough display port

Issue #1216, opened 2024-10-07
Placeholder Text Gets Truncated When Using Japanese Characters


emikolocolv:
Hello,

I've encountered an issue when setting the placeholder text in Japanese (e.g., "エリアを選択してください") using choices.js. The generated input element looks like this:

```html
<input type="search" class="choices__input choices__input--cloned" autocomplete="off" autocapitalize="off" spellcheck="false" role="textbox" aria-autocomplete="list" aria-label="エリアを選択してください" placeholder="エリアを選択してください" style="min-width: 13ch; width: 1ch;">

```

The problem is that the `min-width` is set to 13ch, which causes the placeholder text to be truncated. It would be helpful if there was an option to set the width of the `input[type=search]` or to customize it in some way.

I'm not entirely sure how this is implemented, and I would appreciate any assistance or guidance on how to address this issue.

Thank you for your help!

Xon:
`"エリアを選択してください".length` returns a length of 12 characters, and the `ch` unit assumes non-variable width fonts which is probably what is wrong.

Monkey patching `instance.input.setWidth` may be the best solution.

The original definition is this;
```typescript
  setWidth(): void {
    // Resize input to contents or placeholder
    const { element } = this;
    element.style.minWidth = `${element.placeholder.length + 1}ch`;
    element.style.width = `${element.value.length + 1}ch`;
  }
```

PR #1384, opened 2026-03-06
Fix non-fixed width font support when computing width for input element

terminalchai:
## Problem

The `setWidth()` method in `src/scripts/components/input.ts` computes the search input's `min-width` and `width` style using `str.length + 1` as the number of `ch` units:

```typescript
element.style.minWidth = `${element.placeholder.length + 1}ch`;
element.style.width    = `${element.value.length + 1}ch`;
```

This is wrong for **wide (full-width) Unicode characters** — Japanese (Hiragana/Katakana), Chinese (CJK Ideographs), Korean (Hangul), and fullwidth Latin forms. Each visually occupies **2ch** in a standard browser font, but `.length` counts it as 1.

**Result:** a Japanese placeholder like `エリアを選択してください` (12 codepoints → 24ch wide) is given only 13ch of `min-width`, causing the text to be visibly truncated.

Reported in #1216.

## Steps to Reproduce

```js
new Choices(selectEl, {
  placeholderValue: 'エリアを選択してください',
});
```

The placeholder inside the cloned `<input type="search">` is cut off.

## Solution

Add a module-level helper `getStringWidth(str)` that iterates Unicode code points and returns 2 for characters in wide/fullwidth Unicode blocks (standard *wcwidth*-style logic) and 1 for everything else. Update both `.length` calls in `setWidth()` to use this helper.

Affected Unicode blocks (among others):
- Hiragana / Katakana (U+3041–U+33BF)
- CJK Unified Ideographs (U+4E00–U+9FFF) and Extensions A–E
- Hangul Syllables (U+AC00–U+D7AF)
- Fullwidth Latin / Punctuation (U+FF01–U+FF60)

No behaviour change for pure ASCII input.


src/scripts/components/input.ts
```diff
@@ -1,6 +1,56 @@
...
+ *
+ * Falls back to `str.length` on environments without `Intl.Segmenter`.
+ * For accurate CJK width in those environments, add a polyfill:
+ *   - https://github.com/cometkim/unicode-segmenter
+ *   - https://github.com/surferseo/intl-segmenter-polyfill
+ */
+function getStringWidth(str: string): number {
+  if (typeof Intl !== 'undefined' && typeof Intl.Segmenter === 'function') {
+    let width = 0;
+    for (const { segment } of new Intl.Segmenter().segment(str)) {
```
Xon: This can be replaced with iterating over `[...new Intl.Segmenter().segment(str)]` which is a much lighter weight construct.

Xon (dismissed):
Please ensure `npm run lint` passes.

I'll investigate why the e2e test are failing but it is likely requiring updates for playwright again.

PR #1418, opened 2026-07-19
Focus input synchronously to open iOS keyboard [#1373]

### LINKED ISSUE #1373: Input focus does not trigger keyboard on iOS when showDropdown is opened
On iOS (Safari and iOS WebView), focusing an input element does not open the virtual keyboard when the input is focused as part of opening showDropdown.

As a result, the input appears focused visually, but the keyboard does not open and the user cannot start typing until they manually tap the input.

The same result you can see on demo page

Is there any fix?

AysajanE:
## Summary

- Fix #1373 by making the dropdown visible and focusing its search input synchronously while `showDropdown()` is still in the user-interaction path.
- Keep dropdown measurement, outer-container opening, event dispatch, and selected-choice scrolling deferred to `requestAnimationFrame`.
- Preserve `preventInputFocus` behavior.
- Add a behavior-level Playwright test that proves the input is focused before the deferred layout frame, then verifies the open state after that frame.

Fixes #1373

## Why

On affected iOS Safari and WebView versions, deferring `input.focus()` into `requestAnimationFrame` can move it outside the WebKit user-interaction context needed to show the software keyboard. Because the search input is inside the hidden dropdown, the dropdown must become visible before that synchronous focus call. The surrounding layout work can remain deferred.

## Testing

- `npm run test:unit && npm run lint:js` — passed on the exact head in a network-off clean verifier: 353 unit tests plus JavaScript lint.
- `CI=1 npx playwright test --project=chromium` — passed locally: 222 tests.
- Targeted WebKit Playwright checks for the new focus timing and the two affected paste flows — passed locally: 3 tests.
- Manual iOS-device testing was not performed.

---
AI assistance was used. This change was reviewed by Northset, and I accept responsibility for this submission.

### Verification

[Northset proof-of-pass receipt M-1014](https://northset-oss.github.io/verification-pilot/receipts/M-1014/)  
Contributor self-run; not maintainer verification.

### DIFF
--- a/public/test/select-one/index.html
+++ b/public/test/select-one/index.html
@@ -71,6 +71,7 @@
               const choicesBasic = new Choices('#choices-basic', {
                 allowHTML: false,
               });
+              window.choicesBasic = choicesBasic;
               document
                 .querySelector('button.disable')
                 .addEventListener('click', () => {
--- a/src/scripts/choices.ts
+++ b/src/scripts/choices.ts
@@ -511,14 +511,14 @@ class Choices {
       preventInputFocus = !this._canSearch;
     }
 
+    if (!preventInputFocus) {
+      this.input.focus();
+    }
+
     requestAnimationFrame(() => {
       this.dropdown.show();
       const rect = this.dropdown.element.getBoundingClientRect();
       this.containerOuter.open(rect.bottom, rect.height);
-
-      if (!preventInputFocus) {
-        this.input.focus();
-      }
 
       this.passedElement.triggerEvent(EventType.showDropdown);
 
--- a/src/scripts/choices.ts
+++ b/src/scripts/choices.ts
@@ -511,12 +511,13 @@ class Choices {
       preventInputFocus = !this._canSearch;
     }
 
+    this.dropdown.show();
+
     if (!preventInputFocus) {
       this.input.focus();
     }
 
     requestAnimationFrame(() => {
-      this.dropdown.show();
       const rect = this.dropdown.element.getBoundingClientRect();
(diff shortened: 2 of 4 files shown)


Xon:
@AysajanE can you please give a reason on why this change is required? I'm guessing it is an accessibility thing but can you update the commit message with the why?

AysajanE:
Thanks for asking. This addresses #1373: on affected iOS Safari/WebView versions, moving `input.focus()` into `requestAnimationFrame` can lose the user-interaction context WebKit needs to open the software keyboard. The input then appears focused but requires another tap before typing.

I’ve amended the commit message and PR description with that rationale and the successful test results. I have not claimed a manual iOS-device test.

Xon:
@AysajanE can you also update the e2e tests to verify the change in behavior?

AysajanE:
@Xon I am on it. Sorry I was traveling. I will address it this morning.

AysajanE:
Addressed in `6cab97d`. I added behavior-level Playwright coverage that calls `showDropdown()` and verifies, before the deferred animation frame, that the search input is focused while the outer container is still unopened; after advancing the frame, it verifies the open state. The implementation now makes the hidden dropdown focusable before the synchronous focus call while keeping measurement and outer-container layout deferred. All 353 unit tests and JavaScript lint pass, the full Chromium Playwright suite passes (222/222), and the targeted WebKit checks pass (3/3). Manual iOS-device testing was not performed.

Xon:
LGTM, I plan todo some further testing before making a v11.2.4 release

